PR #10, opened 2018-08-02
Make RETURN an expression.

ml86:
After this change all nodes which can appear in a method body are
expressions which makes some step handling in our DSL easier.

This might seem counter intuitive but it acutally makes sense to
represent RETURN as an expression which has the eval type "Nothing"
which is at the moment represented through the lack of an
EVAL_TYPE edge. Compare to languages like Scala which do the same.


fabsx00:
I agree, that makes sense. While writing the docs, I actually thought it's a bit weird that everything is an expression, and then there are those RETURN nodes that aren't. The docs gloss over this a bit, so, no change required there.

PR #45, opened 2018-10-16
Do not require frontends to draw REF edges to MEMBER nodes.

ml86:


mpollmeier:
fyi: cherry-picked onto master

PR #46, opened 2018-10-22
Print more information in case of error.

ml86:


mpollmeier:
cherry-picked onto master

PR #48, opened 2018-10-22
Add CONTAINED_NODE so that CONTAINED is not reused.

fabsx00:


mpollmeier:
Hang on, this isn't just a rename, it also reverses the edge direction, right?
old: `parent --- "CONTAINS" --> child`
new: `parent <-- "CONTAINED_IN" --- child`

Is this intended? if you want to keep the directions as is, we could name them e.g. `CONTAINS_NODE`

mpollmeier:
Also, id 31 is already in use for `BLOCK` - I'll push my proposal onto this branch.

mpollmeier:
@fabsx00 let me know if that suits, then either you or I can update query-primitives

fabsx00:
This works. Just tested and made adaptions accordingly. Thanks!

mpollmeier:
Cleanup: should we close https://github.com/ShiftLeftSecurity/codescience/issues/1554 and revert https://github.com/ShiftLeftSecurity/codescience/pull/1549 then?

fabsx00:
Good point. Let me do that in a bit.

fabsx00:
There are still issues with https://github.com/ShiftLeftSecurity/codescience/issues/1554, so I am leaving it open. https://github.com/ShiftLeftSecurity/codescience/pull/1549 can indeed be reverted.

PR #53, opened 2018-10-29
support tinkergraph indices

mpollmeier:
based on https://github.com/ShiftLeftSecurity/tinkergraph-gremlin/pull/9


mpollmeier:
I tested with cpg-public, cpg-internal and codescience (some minor patches were required)

PR #55, opened 2018-11-02
cherry-pick changes from tinkergraph branch

mpollmeier:
makes schemas between master and tinkergraph equal again

### DIFF
--- a/codepropertygraph/src/main/resources/schemas/base.json
+++ b/codepropertygraph/src/main/resources/schemas/base.json
@@ -83,7 +83,7 @@
          "comment" : "A method/function/procedure",
          "is": ["DECLARATION"],
          "outEdges" : [
-             {"edgeName": "AST", "inNodes": ["METHOD_RETURN", "METHOD_PARAMETER_IN", "LOCAL",
+             {"edgeName": "AST", "inNodes": ["METHOD_RETURN", "METHOD_PARAMETER_IN",
                                              "MODIFIER", "BLOCK", "TYPE_PARAMETER"]},
              {"edgeName": "CFG", "inNodes": ["CALL", "IDENTIFIER", "LITERAL", "METHOD_REF", "RETURN", "UNKNOWN"]}
          ]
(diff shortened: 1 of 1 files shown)


codepropertygraph/src/main/resources/schemas/base.json
```diff
@@ -255,7 +255,8 @@
         {"id" : 3, "name" : "AST", "comment" : "Syntax child" , "keys" : [] },
         {"id" : 19, "name" : "CFG", "comment" : "Control flow", "keys" : [] },
 
-        {"id" : 9, "name" : "CONTAINS", "keys" : ["LOCAL_NAME", "INDEX"], "comment" : "Shortcut to contained node"},
+        {"id" : 9, "name" : "CONTAINS_NODE", "keys" : ["LOCAL_NAME", "INDEX"], "comment" : "Edge from container to contained node"},
+        {"id" : 28, "name" : "CONTAINS", "keys" : ["LOCAL_NAME", "INDEX"], "comment" : "Shortcut to contained node"},
```
ml86: I am pretty sure the keys have just been introduced as the already existing CONTAINS edge was hijacked for something else. So please reconfirm and than remove those here and on internal branch.

mpollmeier: Hmm, for some reason I thought there was a difference in cardinality, but can't see where that came from. Detangling...

mpollmeier:
I just went through where we use each one, and I'm a bit undecided if it should be one or two separate edges. If we have two, we should do better with the properties and comments. Here's the status quo:

* CONTAINS:
  * {"id" : 28, "name" : "CONTAINS", "keys" : ["LOCAL_NAME", "INDEX"]}
  * TYPE_DECL --- CONTAINS --> METHOD
  * FILE --- CONTAINS --> [TYPE_DECL|METHOD]
  * METHOD --- CONTAINS --> [CALL|IDENTIFIER|LITERAL|RETURN|METHOD_REF|UNKNOWN]
  * created by codescience/cpgenhancements/ContainsEdgeCalculator
  * doesn't look like the keys (localName, index) are ever set or used

* CONTAINS_NODE:
  * {"id" : 9, "name" : "CONTAINS_NODE", "keys" : ["LOCAL_NAME", "INDEX"]}
  * every nodeType automatically get's an outgoing edge of this kind:
    [*] --- "CONTAINS_NODE" --> [*]
  * created by codescience/NewNodeSteps for each NewNode.ContainedNode


Semantically they are very similar, however not identical. The keys are only used for CONTAINS_NODE. We could keep them separate and remove the keys from CONTAINS or make them one (only called CONTAINS I guess). 

I'm leaning towards keeping them separate, mostly because for CONTAINS we know the incoming node type, while for CONTAINS_NODE it can by anything. But that's not a strong opinion. 
@fabsx00  @ml86 what do you think?

ml86:
I strongly think that we should keep those separate as they have two different meanings.
The `CONTAINS ` is a shortcut over multiple `AST` edges.
The `CONTAINS_NODE` is a member ship relation for a compound object.

mpollmeier:
Ok perfect, will do. Your comment (https://github.com/ShiftLeftSecurity/codepropertygraph/pull/55#discussion_r230327005) sounded like you preferred to only have a single one.

mpollmeier:
@ml86 I updated the comments and properties and left it as two separate edgeTypes. Please re-review.

mpollmeier:
equivalent for tinkergraph branch: https://github.com/ShiftLeftSecurity/codepropertygraph/pull/60

PR #59, opened 2018-11-06
Move LOCAL node in AST below BLOCK.

ml86:
This allows use to represent nested local variables with the need for
name mangling.


mpollmeier:
equivalent for master branch: https://github.com/ShiftLeftSecurity/codepropertygraph/pull/61

PR #63, opened 2018-11-07
Move LOCAL node in AST below BLOCK.

ml86:
This allows use to represent nested local variables with the need for name
mangling.

Add 'FUZZYC' as langauge.


ml86:
Wrong base branch.

PR #64, opened 2018-11-18
remove need for tinkergraph-internal branch

mpollmeier:


enhancements/src/main/scala/io/shiftleft/passes/linking/memberaccesslinker/MemberAccessLinker.scala
```diff
@@ -48,7 +48,7 @@ class MemberAccessLinker(graph: ScalaGraph) extends CpgPass(graph) {
                 dstGraph.addEdgeInOriginal(call.underlying, member.underlying, EdgeTypes.REF)
                 finished = true
               case None =>
-                val baseTypes = typ.start.baseType.toList
+                val baseTypes = new NodeTypeDeco(typ).start.baseType.toList
```
fabsx00: In which context do we need this `NodeTypeDeco`? Will this be required on the repl? If so, then I don't think we can take this change.

mpollmeier: thanks for reminding me, I just took it out again. that was only necessarily in some previous state of this branch

fabsx00: Ok, nice.

PR #65, opened 2018-11-19
publish generated sources in `-sources.jar`

mpollmeier:
also ensure they're in the correct directories

note: please review, but I'll only merge together with adaptations to cpg-internal


mpollmeier:
merging together with https://github.com/ShiftLeftSecurity/codepropertygraph-internal/pull/205

PR #80, opened 2018-12-13
Update version of protobuf

vlad-alexandru-ionescu:


Ferada:
Tentative yes, but I'm having problems building `codepropertygraph-internal` right now, plus @mpollmeier is in the middle of a rather large change on this.

vlad-alexandru-ionescu:
Ok, this can wait a while longer

PR #101, opened 2019-04-02
unify public and private projects

mpollmeier:
TODOs that will be handled in follow-up PRs:
* script/main class that runs j2c to produce cpgs for tests
* update example project (once we have a new cpgVersion)


mpollmeier:
we still have one failing test. @b-64rry can hopefully help, please speak to @ml86 for context

PR #114, opened 2019-04-10
upgrade protobuf version to 3.7.0

mpollmeier:
when build configures 3.6.1, local protoc *must be* 3.6.1 (or lower).
it fails with protoc 3.7.0

when build configures 3.7.0, local protoc can be either 3.7.0 or 3.6.1 (or lower).


Ferada:
Can you make the same change for the [`proto` repository](https://github.com/ShiftLeftSecurity/proto) so that they're in sync?

PR #118, opened 2019-04-12
use `nextChecked` instead of `next`, so we get stack traces

mpollmeier:
includes some other iterator-related refactorings

re https://github.com/ShiftLeftSecurity/product/issues/2405


query-primitives/src/main/java/io/shiftleft/cpgloading/ProtoToCpg.java
```diff
@@ -80,8 +81,11 @@ public void addEdges(List<Edge> protoEdges) {
     for (Edge edge : protoEdges) {
       long srcNodeId = edge.getSrc();
       long dstNodeId = edge.getDst();
-      Vertex srcVertex = tinkerGraph.vertices(srcNodeId).next();
-      Vertex dstVertex = tinkerGraph.vertices(dstNodeId).next();
+      if (srcNodeId == -1 || dstNodeId == -1) {
+        throw new IllegalArgumentException("edge " + edge + "has illegal src|dst node. something seems wrong with the cpg");
```
Ferada:
```suggestion
        throw new IllegalArgumentException("edge " + edge + " has illegal src|dst node. something seems wrong with the cpg");
```

PR #128, opened 2019-04-17
Pass in cfgNode implementation via implicit var.

ml86:
This makes it possible to extend functionality in downstream projects.
Include NewXXX classes in NodeVisitor pattern.


ml86:
Not needed for now.

PR #133, opened 2019-04-25
Load argument definition semantics from file instead of hard coding it.

ml86:


fabsx00:
@b-64rry could you write a few sentences about this feature once it works? Otherwise nobody knows about it :)

b-64rry:
> @b-64rry could you write a few sentences about this feature once it works? Otherwise nobody knows about it :)

sure :)

PR #137, opened 2019-04-30
Possible fix for crash URI-related crash on windows

fabsx00:
URLs such as `jar:file:C:/...` are invalid. With this PR, it should be `jar:file://$relativePath`.


fabsx00:
Confirmed to not break on Linux. Safe to merge at least.

PR #140, opened 2019-04-30
trigger build

mpollmeier:


vlad-alexandru-ionescu:
Hmm.. we should get you access to Jenkins @mpollmeier . Can you file a request ticket in `itsupport` repo and assign to Davy?

mpollmeier:
@vlad-alexandru-ionescu this is built on travis.ci, where I can trigger a build, but only an actual commit will trigger a release. my preference would be that I can push directly to master, and I'd argue that's ok because I often do release-specific changes that can't be tested on a branch anyway. 
thanks for approving!

vlad-alexandru-ionescu:
Got it - forgot about that :-)

PR #141, opened 2019-05-02
Make the empty string the default for string fields

fabsx00:


project/DomainClassCreator.scala
```diff
@@ -694,6 +694,7 @@ object DomainClassCreator {
             /* a little hack to allow us to have different DomainClassCreator while using the same cpg traversals */
             if (getHigherType(key) == HigherValueType.Option) " = None"
             else if (key.valueType == "int") " = -1"
+            else if (key.valueType == "String") " = \"\""
```
mpollmeier: slightly nicer to read would be 
```
else if (key.valueType == "String") """ = " """
```

PR #147, opened 2019-05-05
Example for node enrichment

fabsx00:
This PR demonstrates how we can enrich nodes such that enrichments show up in scaladoc - following the same pattern that @mpollmeier used for step enrichment. The mechanism is demonstrated by adding a `parameter` method to `method` nodes.

I realize @ml86 is working on new concepts for our DSL implementation, however, I don't want to wait for that to conclude to make further language improvements. Instead, let's shape the DSL further with the current implementation, and thereby uncover what a new implementation would need to provide.


fabsx00:
I'm ok with that. The broader idea of this PR is, however, to get some feedback on whether this is the correct way to proceed in cases where we need operations that can be performed both on the pipe and the single node. I'm actually hoping for a better suggestion - as an alternative to changing the DSL entirely.

fabsx00:
I'll close this one, and we continue the discussion in https://github.com/ShiftLeftSecurity/codepropertygraph/pull/148

PR #148, opened 2019-05-07
WIP: New DSL implementation proposal.

ml86:
Benefits:
  - The DSL elements are definied once and can be used on single elements
    and pipes of elements.
  - No memory allocation for traversal class. All actions are done non lazy.
  - Get rid of gremlin with all its drawbacks: Being slow, swallowing exceptions etc...


fabsx00:
The key advantage I see here is that there is less boilerplate code to write in order to obtain methods that work both on single nodes and pipes. The implementation also makes very little use of Gremlin (essentially just the Vertex/Edge types and basic graph expansions), removing a source of performance WTFs. On the downside, we don't know if runtime performance of this DSL implementation will be higher, and it's hard to find without investing a considerable amount of additional work.

The language - minus a few details - remains the same. Regardless of whether we go this route or not, now is a good time to add more language specification in the form of tests. I would suggest we hosts these integration tests as samples in the `joern` repository, as they can then also serve as introductory material for people who want to learn the language.

We lose a few language features which we inherited from Gremlin. In particular, this proposal gets rid of as/select. We need to make a conscious design decision here: are these language features worth preserving or have they reached their end of lifetime?

I see this as an improvement. The main question to me is whether this justifies a somewhat major change. I do feel that the ability to easily expand the DSL is a key feature, and improving is quality is worth a time investment. There is a potential for breakage here though, so I would only feel comfortable making these changes if we first create a detailed language spec in the form of tests. That part we need to do anyway.

fabsx00:
Another aspect to investigate: we want to, probably by default, stream nodes through queries to avoid taking up more memory than required. Before we make this move, we should investigate whether it works well with streams, and possibly even parallel streams.

mpollmeier:
I'm not convinced (yet). On the one side we only use a small subset of gremlin, so rebuilding the part we actually need looks fairly straightforward. On the other hand it's a large operation to replace everything we have. We could (and should) do it incrementally, but the total time investment will be significant. 

Here's a few points, ordered from least to most important (IMO). I wasn't able to actually use the new DSL proposal yet because it currently doesn't compile, so these are rather high level. 

1) Exceptions are only an issue with tinkerpop when using the graph api directly, as opposed to using a pipeline. While it's annoying, we did find a simple workaround: `s/.next/.nextChecked/`

2) We only use a small subset of the steps gremlin, and while `as/select` is really cool, we only use is in one place (`CpgMerger`) and don't really need to. I thought about dropping that feature with the current gremlin-scala based dsl, because it would simplify our pipeline types because we don't need all those `HList` type parameters. 

3) Streaming: fabs asked about this, I don't think it'll be an issue

4) Less dependencies: I first imagined it'll be a *lot*, but looking at deps.png (bottom of this post, generated for `ocular/replCore`; `gremlin-core` is in the top-right section) it's actually not that many, and they sound rather small - but I haven't actually checked their sizes. Also note that that's only true once we ported everything and our fork of tinkergraph becomes independent of tinkerpop as well. 

5) Lazyness: From PR description: `All actions are done non lazy`: can you elaborate why this is a good thing?

6) Being able to use all steps on the pipeline and individual nodes: I believe we could achieve this with the current model as well, but I'm not yet convinced that that's a good thing. A node and a traversal are conceptionally different, and calling `.start` to explicitly lift a node into a traversal is the right thing to do IMO. Trying to pretend they are the same thing will lead to confusion and increase complexity (also by relying on even more implicits), irrespective if we do it with the existing or new dsl. 
For every property and relationship defined by the schema in codepropertygraph (e.g. `method.name` and `method.taggedBy`), we can already do this (caveat: `Method.taggedBy` is currently a `List[Id]`, but it'll shortly become `List[nodes.TaggedBy]`), as those are defined in the generated nodes directly. 
Using the Tinkerpop API is cumbersome in certain areas (e.g. in TrackingPoint), and in those cases we can just work directly with the generated nodes instead. The new `VertexRef` wrappers will make that a nicer experience. 

7) Performance: the current branch doesn't compile, but I wanted to get a rough idea what the overhead of the tinkerpop pipeline is, so I compared it to a simple iterator. Here's the results from my laptop for traversing the whole graph:
* iterate all 52k vertices: gremlin-scala 6ms, iterator 3ms
* _also_ follow out edge: gremlin-scala 145ms, iterator 38ms
* _also_ start a new traversal at every vertex: gremlin-scala 550ms, iterator 145ms

The main performance killer is (as expected) to instantiate a new traversal when calling `.start`. I haven't checked what exactly is taking up the time, but I measured that we call it 8k times when running cpg2sp on HSL. The above measurement is for the whole graph, i.e. running `v.start.out.toList` 52k times takes 550ms. **Which means that we're talking about 84ms total time in cpg2sp, which is negligible given the 60s total time for HSL.** The relative changes are significant, but compared to the total time of cpg2sp, it seems like the absolute time spent traversing is only a very small portion. That said, I have only done some rough benchmarks (see code snippet below) - more benchmarking would need to be done to make a more precise statement. 

Conclusion: I'm not convinced yet. The upsides don't feel strong enough to justify the time investment, and in areas where the tinkerpop api isn't a good fit we can directly use the generated nodes. That being said, the PR description is quite brief and I may be missing or misinterpreting some of the benefits. 

Here's the benchmarking code so you can try for yourself with `ocular.sh`. Once the new DSL compiles you can test it with the same code. 
```scala
loadCpg("hello-shiftleft-0.0.1-SNAPSHOT.jar")
val graph = cpg.graph //52k vertices, 134k edges

def timed(msg: String)(fun: () => Unit): Unit = {
  val iterations = 10
  val start = System.nanoTime
  (1.to(iterations)).foreach {_ => fun()}
  val average = (System.nanoTime - start) / iterations.toFloat / 1000000f
  println(s"$msg: ${average}ms")
}

timed("simple iterate: iterator") { () => 
  val iter = graph.vertices()
  while (iter.hasNext) iter.next
} // hsl: 3ms

timed("simple iterate: gremlin-scala") { () => 
  val g = cpg.graph.asScala.traversal
  g.V.iterate
} // hsl: 6ms

timed("V.out: iterator") { () => 
  val res = new java.util.ArrayList[Vertex]
  val iter = graph.vertices()
  while (iter.hasNext) {
    iter.next.vertices(org.apache.tinkerpop.gremlin.structure.Direction.OUT)
  }
} // hsl: 38ms

timed("V.out: gremlin-scala") { () => 
  val g = cpg.graph.asScala.traversal
  g.V.out.toList
} // hsl: 145ms

timed("vertex.start.out: gremlin-scala") { () => 
  val g = cpg.graph.asScala.traversal
  g.V.toList.flatMap { v => 
    v.start.out.toList
  }
} // hsl: 550ms
```

And here's the deps.png ocular dependency graph:
deps

ml86:
Let me clarify/respond to the individual points:
1. You are referencing to the problem of exceptions without a call stack which we solved with `nextChecked`. I was talking about the fact that if you write a traveral like `cpg.method.sideEffect {someCode}.l` and you throw a NoSuchElementException in someCode, gremlin like silently catch it. More important is that is will also screw up its internal logic because it relies on those exceptions for internal control flow. So in theory we would need to wrap functions given to `sideEffect` with a function catching the NoSuchElementException and mapping it to e.g. RuntimeException. This lessens the burden on the DSL user but it still sucks since we are mapping exceptions which make the API description even more complicated.

2. I also think we should drop `as/select` regardless on how we implement the DSL. In my opinion it encourages writing of hard to read queries because you are jumping back and forth between node sets just to produce a one liner. Instead the user should rather work on tuples where that is necessary or split up the query in multiple individual ones. As for the working on tuples: This is in the current DSL very ugly. Compare the old vs new:
Old: `cpg.method.map(m => (m, m.start.methodReturn.head)).l`
New `cpg.method.map(m => (m, m.methodReturn)).l`

3. Stream works and we should do it be changing the `List` in RealPipe to `Stream`. I already tested that locally, works fine.

4. I would see it as a benefit that we control the new DSL completely and have something tailored to our needs. The number and size of our third party dependencies i deem not that important.

5. Sorry for being so confusing here. I wanted to state that we can choose being lazy or not by swapping the underlying implementation in `RealPipe`.

6. I strongly disagree here and think that we need to have the same DSL primitives on individual elements as well as on a pipe of elements for two reasons:
6.1. Imposing a different API for the same thing just because you are working on one instead of many is just cumbersome. All the information whether you are working on one or many would still be there (reflected in the static types). Just look at my example under 2.  I think its clearly an improvement.
6.2. Performance. My proposal avoids any memory allocations inside the DSL mechanics while doing one to one mappings. So by not requiring a lifting of single elements into a traversal we safe a lot of time.

7. I do not have performance measurements for my new DSL proposal yet. But i think we do not need to measure gremlin again. We already know that it sucks if you do millions of x.start.expand.head.

mpollmeier:
1) Exceptions: ah, thanks for clarifying. Yes I agree, that's a problematic side of tinkerpop. Here's an example:
```scala
cpg.method.raw.sideEffect(_ => Nil.head).l
res: List[Method] = List() // should rather throw an Exception
```

2) Dropping as/select: ok as a first step to simplify the codebase I'll drop that, should mostly be a task for `sed`.
'Working with tuples' boils down to 6)
6) Auto-lifting: `m.methodReturn` does indeed look nicer than `m.start.methodReturn.head`, and as layed out before there's two aspects:
* lifting (automatically) calling `.start`: the reason I'm still doubtful that this is not a good idea is not that it's 'one vs many', but because we are lifting into a different level of abstraction: node vs pipeline. Anyway, the implementation with the old dsl would be straightforward:
```scala
implicit def toMethodSteps(node: nodes.Method): io.shiftleft.queryprimitives.steps.types.structure.Method[shapeless.HNil] = node.start
```
* automatically returning the first element (`.head`): I'll comment on this tomorrow after I've had some time playing with it (this branch compiles now).

6.1/7) Performance: 
> So by not requiring a lifting of single elements into a traversal we safe a lot of time.
> But i think we do not need to measure gremlin again. We already know that it sucks if you do millions of x.start.expand.head.
Gremlin does allocate a lot of instances, but if my above benchmark is correct then we're talking about 84ms out of a total 60s cpg2sp run. So yeah it's totally bad, but no it doesn't really make a difference. Again, my benchmark was very high level and may well be missing something.

I also prefer a smaller stack that we control, so we get less surprises as with the exception handling. 
Tomorrow I'll give it a spin on the repl, which will help me understand the auto-expansion of `.head` better.

fabsx00:
We're postponing this, but should revisit. I'm closing this for now. Let's re-open it when we re-open the discussion.

PR #153, opened 2019-05-07
Allow semantics filename to be passed in from outside

fabsx00:
Outside of `cpgqueryingtests`, semantics may be located in a different place, so, with this PR, we can specify its location from the outside.


cpgqueryingtests/src/main/scala/io/shiftleft/cpgqueryingtests/codepropertygraph/CpgFactory.scala
```diff
@@ -6,10 +6,14 @@ import java.nio.file.Files
 import io.shiftleft.cpgloading.{CpgLoader, CpgLoaderConfig}
 import io.shiftleft.queryprimitives.steps.starters.Cpg
 
-class CpgFactory(frontend: LanguageFrontend) {
+class CpgFactory(frontend: LanguageFrontend,
+                 semanticsFilename : String = "cpgqueryingtests/src/test/resources/default.semantics") {
```
ml86: I think it is best to remove the default parameter value here.
It just makes it harder to find the configured file location.

PR #176, opened 2019-05-17
Add first version of a CPG validator.

ml86:
This version only checks the relations between nodes and not
the node properties itself.


cpgvalidator/src/main/scala/io/shiftleft/cpgvalidator/CpgValidator.scala
```diff
@@ -0,0 +1,193 @@
...
+    validateInFacts()
+
+    errorRegistry.getErrorCount == 0
+  }
+
+  def logValidationErrors(): Unit = {
+    errorRegistry.logValidationErrors()
+  }
+
+  private def validateOutFacts(): Unit = {
```
fabsx00: `validateOutFacts` and `validateInFacts` look pretty similar. Any chance we can reduce some code duplication here?

cpgvalidator/src/main/scala/io/shiftleft/cpgvalidator/CpgValidator.scala
```diff
@@ -0,0 +1,193 @@
...
+
+            val allowedEdgeTypes = inFactsByEdgeType.map(_._1)
+            validateAllInEdgesTypes(dstNode, inEdges, allowedEdgeTypes)
+
+          }
+          .iterate()
+    }
+  }
+
+  private def validateOutDegree(srcNode: Vertex, actualDstNodes: List[Vertex], outFact: OutFact): Unit = {
```
fabsx00: Same question for *Degree methods.

cpgvalidator/src/main/scala/io/shiftleft/cpgvalidator/CpgValidator.scala
ml86: Yes can be deduplicated. So far i had no time to do it.

cpgvalidator/src/main/scala/io/shiftleft/cpgvalidator/CpgValidator.scala
ml86: See above.

PR #210, opened 2019-06-04
specify edge ids in protos - avoid id clashes with nodes

mpollmeier:
nodes and edges have the same id namespace - to ensure we don't
accidentally assign an id to an edge that was supposed to be used by a
node later on, we're now storing the edge ids as well

this change is backwards compatible, but all frontends need to be
updated for this to have any effect


mpollmeier:
> nodes and edges have the same id namespace

actually, that's not true! let's hold back on this

PR #216, opened 2019-06-04
WIP CPG splitter

fabsx00:
*Update* since we decided not to merge support for structured CPG archives until we have provided proof of it's value, this PR is now WIP on creation of a CPG splitter which, with little memory usage, splits large CPGs into independent sub CPGs.

---
Previously, we only supported flat cpg.bin.zip files. This PR brings in support for zip files containing directories. We can use directories to bring more structure into the CPG and encode via directory names what is stored in which overlay. This allows us to selectively load overlays per pass and thus reduce the memory footprint.

Fixes that are part of this PR:
- changed signatures of CpgProtoLoader to receive CpgLoaderConfig objects rather than the overflow config only, to allow us to extend loader configuration later without touching this signature again
- Ported CpgProtoLoader to scala to get rid of Scala/Java conversions


fabsx00:
Yep, agreed.

codepropertygraph/src/main/scala/io/shiftleft/codepropertygraph/cpgloading/ProtoCpgLoader.scala
```diff
@@ -0,0 +1,211 @@
...
+      }
+    }
+
+    val filesInDirectory = getFileNamesInDirectory(new File(inputDirectory))
+    // Note: sorting is necessary here because we need to load overlays
+    // in the right order. Sorting requires us to eagerly accumulate
+    // all filenames here. For the directory of overlays, this is not
+    // problematic though, since the number of overlay files in rather small.
+    // If it does ever grow, we may need to rethink this
+    filesInDirectory.sorted(comparator)
```
Ferada:
```suggestion
    filesInDirectory = filesInDirectory.sorted(comparator)
```
Pretty sure you don't have any guarantee that the stream isn't returning the sorted copy instead of in-place sorting?

PR #225, opened 2019-06-12
disable overflow for tests

mpollmeier:
depends on https://github.com/ShiftLeftSecurity/tinkergraph-gremlin/pull/26/

PR #227, opened 2019-06-12
Close resources after use - tests

hubertp:
While running sbt/test continuously I saw a steady increase in the used
heap. This is because CPGs weren't being closed at all and with ondisk
overflow that is actually problematic.

While we might turn off ondisk overflow for tests and such, it is still a
good idea to always close the resources in a clean way.

The change essentially cleans up the way we create/use test fixtures to
ensure that resources are always closed. With the change, running tests
continuously kept heap constant and low.

I refrained from using external library for ARM (such as better files) to keeps things lightweight.


cpgqueryingtests/src/test/scala/io/shiftleft/cpgqueryingtests/steps/CDataFlowTests.scala
```diff
@@ -23,44 +23,45 @@ class CDataFlowTests extends CpgDataFlowTests {
         |     read(fd, buff, sz);
         | }
       """.stripMargin
-    )
+    ).autoClose { cpg =>
```
ml86: I dont get why we need to use `autoClose` here but not for some other tests in other test files.
I would also expect that the cpg is automatically GCed because after this one test case i see no references pointing to it.

hubertp: https://github.com/ShiftLeftSecurity/tinkergraph-gremlin/pull/23

hubertp: Other files are also auto-closed, just explicitly via fixtures. This is the only place where we create Cpgs from code snipets directly and the creation of fixtures is just different.

hubertp: Feel free to run `test` on `master` a couple of times and see how your heap keeps growing

ml86: Ok yes, this other fixtures also close explicitly internally. But this does not explain why we need to do this in the first place if we have no more references to the CPGs.

hubertp: I changed the way we ensure the resources are closed in above tests. Instead of calling `autoclose` explicitly, the closure with CPG ensures this. This certainly simplifies the way tests look and brings only the essential stuff.

Re why we do it at all - see the earlier comment with the PR in tinkegraph. Essentially, not closing the resource means that there are still listeners attached to GC notifications and there are still references that are not being collected. This is certainly counter-intuitive as we no longer use cpg references outside of the scope.
@mpollmeier can probably chip in, if I missed something, but it was the memory leak that we had in a number of places and was visibly causing performance degradation.

ml86: Ok this looks better in terms of not calling `autoClose`.
I still think requiring explicit CPG close is the wrong way to go. @mpollmeier Can we maybe handle this with weak references for your CG notification stuff?

mpollmeier: how would weak references help here? they are always cleared by the next GC, i.e. the overflow feature doesn't work. if you actually meant soft references, those don't help either, since those are cleared when the heap is full - we'd get unpredictable behaviour. 

as of https://github.com/ShiftLeftSecurity/codepropertygraph/pull/225 the overflow feature can be disabled (mostly for tests), which means we don't need to close the cpg any more. if the overflow is used, i don't see a sane way around closing the cpg, and I don't really understand what the problem with that is.

hubertp:
The reason why I'm persistent on having this change in is because CPG is in reality an IO resource. And like any IO resource, such as DB or File, it should be closed cleanly, irrespective of the fact whether it causes a memory leak or not.
Also if you want to test ondisk overflow in tests, then enabling it may cause other problems.

ml86:
@hubertp I think you are right. Since the CPG now has IO semantics (caused by the swap to disk), we need to properly close it.

PR #228, opened 2019-06-12
Minor code nits

hubertp:
Avoiding side-effects and pattern matching whenever not necessary.

### DIFF
--- a/semanticcpg/src/main/scala/io/shiftleft/passes/linking/linker/Linker.scala
+++ b/semanticcpg/src/main/scala/io/shiftleft/passes/linking/linker/Linker.scala
@@ -121,12 +121,12 @@ class Linker(graph: ScalaGraph) extends CpgPass(graph) {
     }
   }
 
-  private def linkToSingle[SRC_NODE_TYPE <: nodes.StoredNode](srcLabels: List[String],
-                                                              dstNodeLabel: String,
-                                                              edgeType: String,
-                                                              dstNodeIdMap: Map[String, JLong],
-                                                              dstFullNameKey: String,
-                                                              dstGraph: DiffGraph): Unit = {
+  private def linkToSingle(srcLabels: List[String],
+                           dstNodeLabel: String,
+                           edgeType: String,
+                           dstNodeIdMap: Map[String, JLong],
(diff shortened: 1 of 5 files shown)


semanticcpg/src/main/scala/io/shiftleft/passes/linking/linker/Linker.scala
```diff
@@ -121,12 +121,12 @@ class Linker(graph: ScalaGraph) extends CpgPass(graph) {
     }
   }
 
-  private def linkToSingle[SRC_NODE_TYPE <: nodes.StoredNode](srcLabels: List[String],
```
ml86: SRC_NODE_TYPE is only temporarily unused until @mpollmeier starts using node types again instead of JLong IDs. So please keep it.

semanticcpg/src/main/scala/io/shiftleft/passes/receiveredges/ReceiverEdgePass.scala
```diff
@@ -31,26 +31,22 @@ class ReceiverEdgePass(graph: ScalaGraph) extends CpgPass(graph) {
     graph.V
       .hasLabel(NodeTypes.CALL)
       .sideEffect { call =>
-        val instanceOption = call
```
ml86: I like breaking up those chains for better readability. Please revert this part.

hubertp: done

semanticcpg/src/main/scala/io/shiftleft/passes/linking/linker/Linker.scala
hubertp: Not really a fan of leaving things because it will be used in some _undefined_ future. And > 2 months since the change doesn't really qualify as _temporarily_ to me.

I would agree with you if this was public interface but it's a private method that can be changed at will.

ml86: Hmm ok.

PR #236, opened 2019-06-21
Introduce TYPE_DECL aliases.

ml86:
Reason: We want to be able to represent C typedef.

Details:
  TYPE_DECL now has a new optional property ALIAS_TYPE_FULL_NAME.
  If this is set the TYPE_DECL is an alias of the provide type.
  Note that is this case the TYPE_DECL cannot provide
  INHERITS_FROM_TYPE_FULL_NAME because deriving and being an
  alias is not possible at the same time. This makes TYPE_DECL
  effectively a sum type with two different characteristics.

Steps:
  Introduced a few new DSL steps to navigate between aliases
  and the referenced types/type declarations.


ml86:
Maybe it did not became clear but the alias TYPE_DECL do not replicate any of the subgraph of the TYPE_DECL they are referring to. The optimal solution in my eyes would have been a TYPE_DECL_INTERFACE, a TYPE_DECL and an ALIAS_TYPE_DECL which implement the interface. But this would be a much bigger change and a hassle in the implementation because our DSL implementation is geared towards sum types and not hierarchies.

ltcmelo:
> Maybe it did not became clear but the alias TYPE_DECL do not replicate any of the subgraph of the TYPE_DECL they are referring to.

@ml86 , in this case, there's even a worse problem in my opinion: we need to have different interpretations of a TYPE_DECL node.

PR #237, opened 2019-06-21
mark slf4j-log4j bridge as Runtime dependency

mpollmeier:


mpollmeier:
because it's scoped as a Runtime dependency, i.e. it shouldn't have any effect on anything downstream: 
```
runtime
This scope indicates that the dependency is not required for compilation, but is for execution. It is in the runtime and test classpaths, but not the compile classpath.
```
https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html

ml86:
Ok, so than please just move the dependency definition in the sub project sbt files.

mpollmeier:
done

PR #240, opened 2019-07-02
Do not extract files from proto zip to file system.

ml86:
Instead we directly work on the proto files in the zip
which saves us the time for copying the files.


mpollmeier:
travis build is flakey, i told @fabsx00  about it. unfortunately due to merge restrictions we can't merge it :(
asked @vlad-alexandru-ionescu to lift those temporarily

PR #258, opened 2019-07-17
Consolidation of files in query-primitives

fabsx00:
I'm turning query-primitives into a package with a clearly identifiable API. This PR begins by consolidating various files and ensuring that we only have to import `io.shiftleft.queryprimitives.steps._` to have all implicits in scope.


query-primitives/src/main/scala/io/shiftleft/queryprimitives/utils/ExpandTo.scala
```diff
@@ -122,20 +105,29 @@ object ExpandTo {
...
+  def reference(node: Vertex): Option[Vertex] =
     node.vertices(Direction.OUT, EdgeTypes.REF).nextOption
-  }
+
+  /**
+    * For a given name, determine whether it is the
+    * name of a "member access" operation, e.g.,
+    * "<operator>.memberAccess".
+    * */
+  def isGenericMemberAccessName(name: String): Boolean =
```
hubertp: That appears like a weird choice for the location of this method.
It is only vaguely related to the other methods which perform the actual expansions.
I'm not saying that `queryprimitives` was better, but at least it didn't have any meaning attached to it.

fabsx00:
The main concerns raised here (most implicits in one file and "where do we put that isGenericMemberAccessName")  apply to the code before and after this PR. The advantage this PR would have brought would have been usability, and that we wouldn't need to explain concepts like "GremlinScalaDeco" to a user. I'll bring in smaller PRs so that we get more focused discussions.

PR #259, opened 2019-07-18
Adjustments due to changes in tinkergraph.

ml86:
Mostly code generator changes.

### DIFF
--- a/codepropertygraph/src/main/scala/io/shiftleft/codepropertygraph/Cpg.scala
+++ b/codepropertygraph/src/main/scala/io/shiftleft/codepropertygraph/Cpg.scala
@@ -35,8 +35,11 @@ object Cpg {
   /**
     * Returns a fresh, empty graph
     */
-  private def emptyGraph: TinkerGraph =
-    TinkerGraph.open(nodes.Factories.AllAsJava, edges.Factories.AllAsJava)
+  private def emptyGraph: TinkerGraph = {
+    val config = TinkerGraph.EMPTY_CONFIGURATION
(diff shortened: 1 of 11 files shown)


codepropertygraph/src/main/scala/io/shiftleft/codepropertygraph/Cpg.scala
```diff
@@ -35,8 +35,11 @@ object Cpg {
   /**
     * Returns a fresh, empty graph
     */
-  private def emptyGraph: TinkerGraph =
-    TinkerGraph.open(nodes.Factories.AllAsJava, edges.Factories.AllAsJava)
+  private def emptyGraph: TinkerGraph = {
+    val config = TinkerGraph.EMPTY_CONFIGURATION
+    config.setProperty(TinkerGraph.GREMLIN_TINKERGRAPH_ONDISK_OVERFLOW_ENABLED, false)
```
mpollmeier: disabling the overflow this will confuse users, and it's working now, so I'd say let's not touch this

mpollmeier: I'm guessing you inherited this from my branch, so just revert that part

mpollmeier: There's no overhead (unless actual serialization is happening of course), and it works fine for small cpgs. 
Markus and I decided to take it out for now and first do some proper testing with big cpgs. The serialization works but can be optimized, which we may want to wait for as well. 
We agreed that @fabsx00 will handle the blame from users, thanks in advance :)

PR #261, opened 2019-07-18
Moved isGenericMemberAccessName into utils package

fabsx00:
There was a single method called `isGenericMemberAccessName` in the package object of query-primitives. This is a utility method used in different passes and in the `ExpandTo` utility class. I have moved this method into a new object in the utils package so that it is in a less prominent place. I also believe we should not expose this beyond query-primitives, but that is a change for another day.

### DIFF
--- a/query-primitives/src/main/scala/io/shiftleft/queryprimitives/utils/MemberAccess.scala
+++ b/query-primitives/src/main/scala/io/shiftleft/queryprimitives/utils/MemberAccess.scala
@@ -1,0 +1,20 @@
+package io.shiftleft.queryprimitives.utils
+
+import io.shiftleft.codepropertygraph.generated.Operators
+
+object MemberAccess {
+
+  /**
+    * For a given name, determine whether it is the
+    * name of a "member access" operation, e.g.,
+    * "<operator>.memberAccess".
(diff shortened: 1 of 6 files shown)


query-primitives/src/main/scala/io/shiftleft/queryprimitives/utils/MemberAccess.scala
```diff
@@ -0,0 +1,20 @@
...
+import io.shiftleft.codepropertygraph.generated.Operators
+
+object MemberAccess {
+
+  /**
+    * For a given name, determine whether it is the
+    * name of a "member access" operation, e.g.,
+    * "<operator>.memberAccess".
+    * */
+  def isGenericMemberAccessName(name: String): Boolean = {
```
hubertp:
```suggestion
  private[queryprimitives] def isGenericMemberAccessName(name: String): Boolean = {
```
if you don't want to expose it outside of `queryprimitives`

fabsx00: Currently, we can't do that yet, but when we can, I will, thx.

PR #267, opened 2019-07-20
Have CpgPass work on CPG, not ScalaGraph

fabsx00:
Before this PR, CpgPasses operate on ScalaGraphs, which is problematic for two reasons:
- we motivate people to work on the raw graph as opposed to using our DSL, and so, any changes in the code property graph layout will break people's passes
- We tie the CPG pass concept to gremlin, which is unnecessary.

This PR therefore changes the signature of CpgPasses such that they receive Cpgs as opposed to the underlying graphs. Of course, if you know what you are doing, you can still access the underlying graph via `cpg.graph` - with all the downsides this brings.

A bit of cleanup of the CpgLoader is also done in this PR. In particular CpgOverlayLoader is made private, so that we can be sure users can only go via the CpgLoader object for all CPG-loading related operations.


semanticcpg/src/test/scala/io/shiftleft/passes/MethodDecoratorPassTests.scala
```diff
@@ -20,7 +21,7 @@ class MethodDecoratorPassTests extends WordSpec with Matchers {
 
     method --- EdgeTypes.AST --> parameterIn
 
-    val methodDecorator = new MethodDecoratorPass(graph)
+    val methodDecorator = new MethodDecoratorPass(new Cpg(graph.graph))
```
mpollmeier: since we call this a lot, we could add the following to the Cpg companion object:
```
def apply(graph: ScalaGraph) = new Cpg(graph.graph)
```
and then replace this line (and many others) with
```
val methodDecorator = new MethodDecoratorPass(Cpg(graph))
```

mpollmeier: n.b. this pattern is used a lot in the CS repo as well, so just adjust the companion PR as well

fabsx00: Good idea. I'll merge this one and then do that in a separate PR.

PR #270, opened 2019-07-23
Rework CpgLoaderConfig

fabsx00:
I am currently documenting `CpgLoader`, and I found that following problems:
- there was no nice way of disabling indices and overflow at the same time, or in general, to take default values and change a subset of the config options
- `ignoredProtoEntries` is not required anymore. This feature was never documented, so I doubt customers are using it.

This PR adds methods to the `CpgLoaderConfig` to allow for a builder-like Config construction. It also removes `ignoredProtoEntries`. The methods `withStorage` and ``withoutStorage` in the `CpgLoaderConfig` object are kept but deprecated.

### DIFF
--- a/codepropertygraph/src/main/scala/io/shiftleft/codepropertygraph/cpgloading/CpgLoaderConfig.scala
+++ b/codepropertygraph/src/main/scala/io/shiftleft/codepropertygraph/cpgloading/CpgLoaderConfig.scala
@@ -2,25 +2,27 @@ package io.shiftleft.codepropertygraph.cpgloading
 
 object CpgLoaderConfig {
 
+  // To add a new option, please add a default value below and corresponding
+  // methods in the case class.
+
   def default =
     CpgLoaderConfig(
       createIndices = true,
       onDiskOverflowConfig = Some(OnDiskOverflowConfig()),
-      ignoredProtoEntries = None
     )
 
+  @deprecated("Use CpgLoaderConfig.default.withStorage instead", "Jul 19")
   def withStorage(path: String) =
     CpgLoaderConfig(
       createIndices = true,
       onDiskOverflowConfig = Some(OnDiskOverflowConfig(graphLocation = Some(path))),
-      ignoredProtoEntries = None
     )
 
+  @deprecated("Use CpgLoaderConfig.default.withoutStorage instead", "Jul 19")
   def withoutOverflow =
     CpgLoaderConfig(
       createIndices = true,
-      onDiskOverflowConfig = None,
-      ignoredProtoEntries = None
+      onDiskOverflowConfig = None
     )
 
 }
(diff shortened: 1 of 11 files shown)


codepropertygraph/src/main/scala/io/shiftleft/codepropertygraph/cpgloading/CpgLoaderConfig.scala
```diff
@@ -2,25 +2,27 @@ package io.shiftleft.codepropertygraph.cpgloading
...
+  // methods in the case class.
+
   def default =
     CpgLoaderConfig(
       createIndices = true,
       onDiskOverflowConfig = Some(OnDiskOverflowConfig()),
-      ignoredProtoEntries = None
     )
 
+  @deprecated("Use CpgLoaderConfig.default.withStorage instead", "Jul 19")
```
hubertp: Adding the deprecation date for the `since` field is a rather odd way, especially since `July 19` is rather ambiguous. Either we don't use that field completely (also in other places) or we should be using a version.

fabsx00: I've removed "July 19". However, if I don't at least pass an empty string, I get a compiler warning.

codepropertygraph/src/main/scala/io/shiftleft/codepropertygraph/cpgloading/CpgLoaderConfig.scala
```diff
@@ -2,34 +2,61 @@ package io.shiftleft.codepropertygraph.cpgloading
...
-                           onDiskOverflowConfig: Option[OnDiskOverflowConfig],
-                           ignoredProtoEntries: Option[IgnoredProtoEntries])
+case class CpgLoaderConfig(createIndexes: Boolean, onDiskOverflowConfig: Option[OnDiskOverflowConfig]) {
+
+  /**
+    * Existing configuration with overflowdb path as specified
+    * @param path the path of the on disk overflow database
+    * */
+  def withStorage(path: String): CpgLoaderConfig =
+    this.copy(onDiskOverflowConfig = Some(OnDiskOverflowConfig(graphLocation = Some(path))))
```
mpollmeier: onDiskOverflowConfig has more options. proposal (may not compile):
```
def withStorage(path: String): CpgLoaderConfig = {
  val newOndiskConfig = onDiskOverflowConfig match {
    case None => OnDiskOverflowConfig(graphLocation = Some(path))
    case Some(old) => old.copy(graphLocation = Some(path))
  }
  this.copy(onDiskOverflowConfig = Some(newOndiskConfig))
}

                               
def withSwappingFromHeapPercentage(heapPercentageThreshold: Int): CpgLoaderConfig = {
  val newOndiskConfig = onDiskOverflowConfig match {
    case None => OnDiskOverflowConfig(heapPercentageThreshold = heapPercentageThreshold)
    case Some(old) => old.copy(heapPercentageThreshold = heapPercentageThreshold)
  }
  this.copy(onDiskOverflowConfig = Some(newOndiskConfig))
}

```

PR #281, opened 2019-07-30
Improvements for AST traversals

fabsx00:
- Fixed `condition` for for-loops. It returned the initializer earlier
- Introduced `inAst` and `inAstMinusLeaf` to walk upwards in the AST
- Ensure that AST nodes always have an "order" property

@tuxology this should allow you the traversal you were asking for.

### DIFF
--- a/query-primitives/src/main/scala/io/shiftleft/queryprimitives/steps/types/expressions/ControlStructure.scala
+++ b/query-primitives/src/main/scala/io/shiftleft/queryprimitives/steps/types/expressions/ControlStructure.scala
@@ -1,6 +1,6 @@
 package io.shiftleft.queryprimitives.steps.types.expressions
 
-import gremlin.scala.GremlinScala
+import gremlin.scala.{BranchCase, BranchOtherwise, GremlinScala}
 import io.shiftleft.codepropertygraph.generated.{NodeKeys, nodes}
 import io.shiftleft.queryprimitives.steps.NodeSteps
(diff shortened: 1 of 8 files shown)


query-primitives/src/main/scala/io/shiftleft/queryprimitives/steps/types/expressions/ControlStructure.scala
```diff
@@ -25,8 +25,23 @@ class ControlStructure[Labels <: HList](raw: GremlinScala.Aux[nodes.ControlStruc
...
+          node.start.astChildren.order(2).headOption
+        } else {
+          node.start.astChildren.order(1).headOption
+        }
+      }.filterOnEnd(_.isDefined).map(_.get).raw.cast[nodes.Expression]
+    )
+  }
+
+  private def isForLoop(node: nodes.AstNode): Boolean =
+    node.start.astChildren.l.size == 4
```
Ferada: Necessarily? You can have `for` loops in some languages with only the body for example.

fabsx00: This currently only works for the fuzzy C frontend. We are missing a node of type "CONDITION" here because without it, there's no way of knowing exactly which of the expressions is the condition.

fabsx00: Just talked to @ml86 about this. We'll introduce a CONDITION edge, to be drawn between the control structure and the condition.

PR #287, opened 2019-07-31
Fix regenerate test cpg script.

ml86:
Previous version required pressing enter for whatever reason.
Also made script less arcane.


mpollmeier:
asking for the java2cpg home was a feature, not a bug. the script made an educated guess, but allowed the user to use a different location.

Issue #280, opened 2019-07-29
Provide hint when git-lfs is not set up


fabsx00:
I reinstalled my system and I didn't have git-lfs installed and had not run git-lfs pull. As a consequence, `sbt test` failed with a rather mysterious message about broken ZIP files. We should catch this error and suggest to install `git-lfs`.

PR #288, opened 2019-08-01
Added git-lfs check to project build file. (#280)

lifecoder:


lifecoder:
> I didn't try it by uninstalling git lfs, but this looks fine.
> Is the `ProcessLogger` necessary?

@mpollmeier It is optional, but without it it just prints list of files under git-lfs to console on every build

ml86:
@mpollmeier I think to remember that you already implemented some kind of warning mechanism like that. Am I right? If yes i think we should remove the old stuff.

mpollmeier:
@ml86 nope I just added a how-to-install to the cs readme: https://github.com/ShiftLeftSecurity/codescience#how-to-setup-git-lfs
we could copy that to this repo - thoughts?

ml86:
@mpollmeier If that is still required please also add this here.

lifecoder:
@mpollmeier We already have similar instructions here on how to run/install git-lfs here just at the beginning of the README

Issue #279, opened 2019-07-29
ProtoCpgLoader.loadOverlays should return Iterator<CpgOverlay>


fabsx00:
Currently `ProtoCpgLoader.loadOverlays` returns a list of CpgOverlays, which means that we need to hold all overlays in memory at once. We should rather return an iterator and ensure that the users of `ProtoCpgLoader.loadOverlays` do not gather all on this iterator. This is probably something we should do after porting ProtoCpgLoader to Scala.

PR #289, opened 2019-08-01
Load overlays through iterator. (#279)

lifecoder:


codepropertygraph/src/main/scala/io/shiftleft/codepropertygraph/cpgloading/CpgOverlayLoader.scala
```diff
@@ -20,9 +22,17 @@ private[cpgloading] object CpgOverlayLoader {
...
-      applier.applyDiff(overlay)
-    }
+    ProtoCpgLoader.loadOverlays(filename)
+      .map { overlays: Iterator[CpgOverlay] =>
+        overlays.foreach(applier.applyDiff)
+      }
+      .tried
+      .recover {
+        case e: IOException =>
+          e.printStackTrace()
```
mpollmeier: please use a logger. this also runs in `mesh` (part of the CS repo), where this exception would get lost

PR #294, opened 2019-08-06
remove test jars from the repository, generate them instead

mpollmeier:
script now generates cpgs from sources, by compiling them and creating
jars, rather than having the jars checked into the repository

it also allows for 'pregenerated' jars like helloshiftleft, since we
don't want to invoke the full maven build


ml86:
So how is somebody who does not have access to java2cpg now running the tests? I think the cpg.bin.zip files need to stay. Just the intermediate *jar can be removed.

mpollmeier:
@ml86 the cpgs are still part of the repo, just in a different path. the jars have been removed

PR #297, opened 2019-08-08
swap TinkerGraph for OverflowDB, simplify codegen

mpollmeier:


mpollmeier:
This is green and using a released version of overflowdb now. I'll prepare the downstream PRs and will merge together.

PR #298, opened 2019-08-09
Cleanup test structure

fabsx00:
This PR is in reference to
https://github.com/ShiftLeftSecurity/codescience/issues/2710

- The project `cpgqueryingtests` - containing tests for semanticcpg, query-primitives and dataflowengine - is dissolved. Its tests are moved to the corresponding projects so that they are in the same project as the code they test.
- The package `query-primitives` is merged into `semanticcpg` because the query language it defines only works after applying the passes in `semanticcpg` and it seems more technical than useful to untangle those parts of the language usable before applying these passes and after, in particular as the passes work on the raw graph and do not use the language.
- Scalafmt was executed. Unfortunately, this formatted some files which are unrelated to this change.
- CpgNodeMethods is simplified (detailed explanation in the commit)
- Test fixtures are consolidated. We had a few of these, which were quite similar.


fabsx00:
Most of these are files that are being moved around. I just went over the PR myself. Please try it, it's reviewable.

fabsx00:
Did scalafmt in an extra PR and rebased.

PR #301, opened 2019-08-12
Added keys validation to cpgvalidator.

max-leuthaeuser:
Refactored cpgvalidator:
 * extracted existing validators
 * extracted facts
 * extracted import logic

Added keys validation and tests for it.


Ferada:
I'm guessing we don't have an automatic check for it, but this has some formatting changes that disagree with `scalafmt` output.

PR #303, opened 2019-08-12
Add scalafmt check to Travis build

hubertp:
Adding a separate linting task that should prevent people from
accidentally pushing non-formatted code.


Ferada:
Can't that be in the same line so `sbt` doesn't have to be started twice?

PR #315, opened 2019-08-19
delete tmp files during compilation

mpollmeier:
re https://github.com/ShiftLeftSecurity/product/issues/2729

### DIFF
--- a/project/DomainClassCreator.scala
+++ b/project/DomainClassCreator.scala
@@ -737,9 +737,7 @@ object DomainClassCreator {
     targetFile.createIfNotExists(createParents = true)
 
     println(s"writing results to $targetFile")
-    outputFile.copyTo(targetFile, overwrite = true)
-    outputFile.delete()
-    targetFile.toJava
+    outputFile.moveTo(targetFile, overwrite = true).toJava
   }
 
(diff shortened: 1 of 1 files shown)


project/DomainClassCreator.scala
```diff
@@ -738,6 +738,7 @@ object DomainClassCreator {
 
     println(s"writing results to $targetFile")
     outputFile.copyTo(targetFile, overwrite = true)
```
Ferada:
```suggestion
    outputFile.moveTo(targetFile, overwrite = true)
```

project/DomainClassCreator.scala
```diff
@@ -738,6 +738,7 @@ object DomainClassCreator {
...
+    outputFile.delete()
```
Ferada:
```suggestion
```

project/DomainClassCreator.scala
Ferada: Unless you want the old file to be there if the process fails between the copy and the delete?

project/DomainClassCreator.scala
```diff
@@ -738,6 +738,7 @@ object DomainClassCreator {
...
     targetFile.toJava
```
Ferada:
```suggestion
```

PR #317, opened 2019-08-21
replace nodevisitor pattern by pattern matches

mpollmeier:
cleaner and more condense code


mpollmeier:
build error is just scalafmt. 
before merging this I'll ensure codescience has a matching PR, but I just wanted to test the waters before doing so.

mpollmeier:
good point about the `lookupswitch|tableswitch` - scalac can indeed not use those, because the match contains type checks. will close this and add a comment to the NodeVisitor code, for the next person wondering.

mpollmeier:
https://github.com/ShiftLeftSecurity/codepropertygraph/pull/318

PR #319, opened 2019-08-22
remove 'special' ext/Enrichable pattern

mpollmeier:
We introduced this mechanism to add implicits without the need for explicit
imports, mostly so that they will appear in scaladoc. This requirement has
been dropped now, so that the downsides (less readable, complex, not
commonly used) outweigh the benefits.


mpollmeier:
superseded by https://github.com/ShiftLeftSecurity/codepropertygraph/pull/328

PR #340, opened 2019-09-03
Intern duplicate string/symbols in the CPG Loader

lifecoder:
Removing duplicate strings frees up to 15% of heap on a moderate size cpg.

### DIFF
--- a/codepropertygraph/src/main/scala/io/shiftleft/codepropertygraph/cpgloading/ProtoToCpg.scala
+++ b/codepropertygraph/src/main/scala/io/shiftleft/codepropertygraph/cpgloading/ProtoToCpg.scala
@@ -1,8 +1,10 @@
 package io.shiftleft.codepropertygraph.cpgloading
 
 import java.lang.{Boolean => JBoolean, Integer => JInt, Long => JLong}
+import java.util
 import java.util.{NoSuchElementException, Collection => JCollection, Iterator => JIterator}
 
+import gnu.trove.map.hash.THashMap
 import io.shiftleft.codepropertygraph.Cpg
 import io.shiftleft.proto.cpg.Cpg.CpgStruct.{Edge, Node}
 import io.shiftleft.proto.cpg.Cpg.PropertyValue
@@ -14,26 +16,30 @@ import io.shiftleft.overflowdb.OdbGraph
 import scala.collection.JavaConverters._
 import scala.collection.mutable.ArrayBuffer
 import io.shiftleft.overflowdb.OdbConfig
+import io.shiftleft.utils.StringInterner
 
 object ProtoToCpg {
   val logger: Logger = LogManager.getLogger(classOf[ProtoToCpg])
 
-  def addProperties(keyValues: ArrayBuffer[AnyRef], name: String, value: PropertyValue): Unit = {
+  def addProperties(keyValues: ArrayBuffer[AnyRef],
+                    name: String,
+                    value: PropertyValue,
+                    interner: StringInterner = StringInterner.noop): Unit = {
     import io.shiftleft.proto.cpg.Cpg.PropertyValue.ValueCase._
     value.getValueCase match {
       case INT_VALUE =>
-        keyValues += name
+        keyValues += interner.intern(name)
         keyValues += (value.getIntValue: JInt)
       case STRING_VALUE =>
-        keyValues += name
-        keyValues += value.getStringValue
+        keyValues += interner.intern(name)
+        keyValues += interner.intern(value.getStringValue)
       case BOOL_VALUE =>
-        keyValues += name
+        keyValues += interner.intern(name)
         keyValues += (value.getBoolValue: JBoolean)
       case STRING_LIST =>
         value.getStringList.getValuesList.asScala.foreach { elem: String =>
-          keyValues += name
-          keyValues += elem
+          keyValues += interner.intern(name)
+          keyValues += interner.intern(elem)
         }
       case VALUE_NOT_SET => ()
       case _ =>
@@ -49,6 +55,7 @@ class ProtoToCpg(overflowConfig: OdbConfig = OdbConfig.withoutOverflow) {
     OdbGraph.open(overflowConfig,
                   io.shiftleft.codepropertygraph.generated.nodes.Factories.AllAsJava,
                   io.shiftleft.codepropertygraph.generated.edges.Factories.AllAsJava)
+  private val interner: StringInterner = StringInterner.makeStrongInterner()
 
   def addNodes(nodes: JCollection[Node]): Unit =
     addNodes(nodes.asScala)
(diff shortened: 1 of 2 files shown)


codepropertygraph/src/main/scala/io/shiftleft/codepropertygraph/cpgloading/ProtoToCpg.scala
```diff
@@ -49,6 +53,7 @@ class ProtoToCpg(overflowConfig: OdbConfig = OdbConfig.withoutOverflow) {
     OdbGraph.open(overflowConfig,
                   io.shiftleft.codepropertygraph.generated.nodes.Factories.AllAsJava,
                   io.shiftleft.codepropertygraph.generated.edges.Factories.AllAsJava)
+  private val interner: StringInterner = StringInterner.makeStrongInterner()
```
mpollmeier: StrongInterner? typo

codepropertygraph/src/main/scala/io/shiftleft/utils/StringInterner.scala
```diff
@@ -0,0 +1,34 @@
...
+}
+
+object StringInterner {
+  val DefaultMaxStringLength: Int = 1024
+  val DefaultInitialSize: Int = 64 * 1024
+
+  val noop: StringInterner = new StringInterner {
+    override def intern(s: String): String = s
+  }
+  def makeStrongInterner(maxStringLength: Int = DefaultMaxStringLength,
```
mpollmeier: why not just a regular class with a constructor? would have avoided the typo as well :)

codepropertygraph/src/main/scala/io/shiftleft/utils/StringInterner.scala
```diff
@@ -0,0 +1,34 @@
+package io.shiftleft.utils
+
+import gnu.trove.map.hash.THashMap
+
+trait StringInterner {
```
mpollmeier: Please add some documentation about what this does and why we have it. 
This is especially important because there are no tests - I wouldn't know how to test this either, so I guess just add docs so that the next person is less likely to throw this out.

mpollmeier:
so everything i read claims that the jvm does this internally as of jdk7, but it has a very low default size for interning string, apparently it's 60013 strings in java8. we can control it with `-XX:StringTableSize` though. still feels like we don't need this. can you repeat your benchmark by specifying a higher stringtablesize instead of using our custom interner please?
this one is a good summary: http://java-performance.info/string-intern-in-java-6-7-8/

codepropertygraph/src/main/scala/io/shiftleft/utils/StringInterner.scala
```diff
@@ -0,0 +1,34 @@
...
+
+  val noop: StringInterner = new StringInterner {
+    override def intern(s: String): String = s
+  }
+  def makeStrongInterner(maxStringLength: Int = DefaultMaxStringLength,
+                         initialSize: Int = DefaultInitialSize): StringInterner = new StringInterner {
+    private val stringCache = new THashMap[String, String](initialSize)
+
+    def intern(s: String): String = {
+      if (s.length < maxStringLength) {
```
ml86: Why do we not intern long strings? Wouldnt we save most memory for those?

ml86:
Actually would also expect the JVM to do this. I fact while analyzing the memory problems we had with the graph layout i have seen in the memory dump that the JVM deduplicates at least a big part of the string automatically. 

After all it is a very much contained change which is easy to understand, so the risk of bringing it in is minimal. I would still like to here same reasoning/explanation why the JVM wont do this for us.

lifecoder:
> Actually would also expect the JVM to do this. I fact while analyzing the memory problems we had with the graph layout i have seen in the memory dump that the JVM deduplicates at least a big part of the string automatically.
> 
> After all it is a very much contained change which is easy to understand, so the risk of bringing it in is minimal. I would still like to here same reasoning/explanation why the JVM wont do this for us.

I know that this looks a bit "hacky" and JVM supposed to do string deduplication by itself when we specify `-J-XX:+UseG1GC -J-XX:+UseStringDeduplication`
I did few more tests to be sure and I can see that JVM does indeed deduplicate strings
when those flags are added
As far as I understand it is not guaranteed to happen unless there is some CPU idle time, but we have some blocking time while doing IO with zip file.
I looked at the stats of the JVM string deduplication while working with medium sized cpg and it does indeed save some space.

But when I provide smaller heap size to force the OOM and we are running processing on the graph I think it does not have the idle CPU time to deduplicate all the strings and I get OOM error and when I inspect the heap dump it has lots of duplicated strings from CPG. When I add the code in this PR it has much less duplicated strings and does not throw OOM error.

I think we should add `-J-XX:+UseG1GC -J-XX:+UseStringDeduplication` to our JVM process args  along with the code in this PR. So we avoid OOMs and at the same time save some additional memory from duplicated strings coming from other sources than CPG.

@ml86 @mpollmeier What do you think about this?

mpollmeier:
You're right, it's only enabled with `-XX:+UseStringDeduplication`, my assumption was that it's on by default. Maybe with later JDKs. 

Another interesting fact: according to https://dzone.com/articles/usestringdeduplication, only the underlying `char[]` are deduplicated, the `String` instances themselves are not. So your StringInterner will give us better results anyway. 

We already use G1GC in the most relevant places (mostly ocular.sh, cpg2sp.sh), and I asked @hubertp to configure it for mesh, but not sure if that was done. Please create separate PRs for ocular/codescience to add `-J-XX:+UseStringDeduplication` to those. 

tl;dr; your plan sounds good to me. Please note my other comments on this PR (code comments)

ml86:
@lifecoder If we cannot rely on the JVM string deduplication i thing we should use your string internalization. So go ahead with this PR.

codepropertygraph/src/main/scala/io/shiftleft/utils/StringInterner.scala
lifecoder: @mpollmeier I've added some docs that should hint why it is "Strong" instead of "String" :)
My plan was to have "WeakInterner" in addition to the strong one, that will hold weak refs for use cases where it is not good to keep refs to all the values inside a hashmap.

mpollmeier: ah, hehe ok. since we only have one version, why not make it a class with a normal constructor, rather than a `make` function? anyway, no strong feelings, you can leave it as is if you prefer.

mpollmeier:
agreed, but don't overlook markus comment re long strings (strongs?)

codepropertygraph/src/main/scala/io/shiftleft/utils/StringInterner.scala
lifecoder: @ml86 Looking at heap dump I've found that almost all of the savings come from small-medium strings, like java method signatures so I've decided to limit somehow the amount of memory that it holds on to.

codepropertygraph/src/main/scala/io/shiftleft/utils/StringInterner.scala
lifecoder: @mpollmeier I have no strong feelings about this but I prefer to have factory method, so when you provide new implementation with different type of hashmap or some other optimization, you can easily change it in one place and try it out.

mpollmeier: ok

PR #341, opened 2019-09-03
Move callgraph support into its own project

fabsx00:
Prior to this PR, `semanticcpg.language` already offers a call graph based on a "call resolver" of type `ICallResolver`. In practice, this resolver does not do anything for the OSS version, and it uses our data flow tracker for points-to analysis in the commercial version. Conceptually, there is a tight integration between call graphs and data flow analysis. Call graphs are needed for interprocedural data flow analysis, and data flow analysis is needed to construct call graphs. It therefore makes little sense to put the dataflowengine into its own project - and see it as an extension of the semantic cpg - but leave the callgraph as part of the semanticcpg.

It first seemed to me that moving the callgraph into `dataflowengine` would be the right approach, however, note that `dataflowengine` in `codepropertygraph` is our OSS data flow engine - which we do not use in `codescience` and Ocular. Had we placed the callgraph here, we would have received a rather artificial dependency of the commercial data flow engine on the OSS data flow engine.

Instead, this PR moves the callgraph into its own project, resulting in a nice isolation of the callgraph-related language elements.


ml86:
I can live with this but i personally feel that the call resolver is not worth a separate sub project.
I agree that putting it in the open source source data flow engine makes because it would create a strange dependency relation.
How about having data flow engine base/interface project and than an open source and close source version building up on this. That way we could put the call resolver in the base part of the data flow engine.

fabsx00:
Yes, I share your sentiment about this project being too small. Now that the call-related query language is isolated, let me see if I can go with the approach of a "base dataflowengine" project instead. I'll continue to work on this branch.

fabsx00:
The way things currently are, there is no shared code between the OSS and the commercial data flow engine, and rather little potential for code sharing. In fact, even the language specified by the OSS data flow engine is different than that of the commercial one. Let me investigate instead if it makes sense to take more related code from `semanticcpg` and host that in `callgraph`.

fabsx00:
There is no more code from `semanticcpg` that deals with call graphs. I have therefore moved `callgraph` back into `semanticcpg`, however, I have kept the call-graph related language elements isolated in extension classes to have the clean separation anyway.

PR #348, opened 2019-09-11
Fix cpgvalidator error reporting

AlexDenisov:
Currently, all the validators inherit from the `Validator` class.
The `Validator` holds a property `errorRegistry`, which is supposed to be
filled in by the children classes. After validation, the base class
prints all the errors found by its children classes. The children
classes, however, get their own copy of the `errorRegistry` and save
errors into it, leaving the `errorRegistry` of the base class empty.
So the base class then prints all the errors from an empty
`errorRegistry`.

This patch makes sure that only one `errorRegistry` exists and is shared
between all the validators.


ml86:
Please also add a test which makes sure that this problem does not occur again.

AlexDenisov:
@ml86 I added a test. While doing that I got a question: do we actually need the `CpgValidator` class at all? It's too tiny, imo.

PR #362, opened 2019-09-24
fix protoc

mpollmeier:
not passing resulting `File` because protoc requires a relative file :(


mpollmeier:
@drewshiftleft it does, thank you!

PR #367, opened 2019-09-27
Implement new CPG server.

GlassAndOneHalf:
- Written from the group up using a typelevel FP stack for maintainability and extensibility.


cpgserver/src/main/scala/io/shiftleft/cpgservernew/route/CpgProvider.scala
```diff
@@ -0,0 +1,31 @@
...
+    */
+  def createCpg(fileNames: Set[String]): IO[UUID]
+
+  /**
+    * Returns a CPG result identified by `uuid` iff the CPG has been created.
+    *
+    * @param cpgId A UUID associated with a CPG.
+    * @return An [[OptionT]], containing a CPG result if it exists.
+    */
+  def retrieveCpg(cpgId: UUID): OptionT[IO, CpgOperationResult[Cpg]]
```
fabsx00: Seems like a misnomer, I think, because one would except an entire CPG to be returned, while really, it just informs us about the current state the CPG is in.

cpgserver/src/main/scala/io/shiftleft/cpgservernew/route/CpgRoute.scala
```diff
@@ -0,0 +1,92 @@
...
+import org.http4s.circe._
+import org.http4s.dsl.io._
+import org.http4s.{EntityDecoder, HttpRoutes}
+
+final class CpgRoute(implicit cpgProvider: CpgProvider,
+                     cpgQueryExecutor: CpgQueryExecutor) {
+
+  import CpgRoute._
+
+  private def allFilesExist(fileNames: List[String]): IO[Boolean] = fileNames.forallM { fileName =>
```
fabsx00: This method seems overlay long. Can we split it up into multiple smaller methods where names indicate higher-level operations?

cpgserver/src/main/scala/io/shiftleft/cpgservernew/route/CpgProvider.scala
GlassAndOneHalf: It's true that currently we only use this to determine whether or not the CPG is ready, however, it could be used by another component of the server that operates over the CPG, I'm trying to keep things in the `CpgProvider` as general as possible.

cpgserver/build.sbt
```diff
@@ -8,7 +7,67 @@ dependsOn(Projects.codepropertygraph)
...
+  "-Ywarn-unused:privates",            // Warn if a private member is unused.
+  "-Ywarn-value-discard"               // Warn when non-Unit expression results are unused.
+)
+
+val ScalatraVersion = "2.6.5"
+
+val Http4sVersion = "0.20.11"
+val CirceVersion = "0.11.1"
 libraryDependencies ++= Seq(
+  "org.http4s" %% "http4s-blaze-server" % Http4sVersion,
```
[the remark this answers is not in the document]
GlassAndOneHalf: There's a library that works with Blaze/http4s that will automatically generate documentation for us called [rho](https://index.scala-lang.org/http4s/rho/rho-core/0.14.0).

As for why I chose Blaze, I believe it is a better framework than Scalatra as it is much more composable and testable owing to the fact it is built in an FP style. This also allows us to easily build additional functionality using the typelevel stack. It also has support for websockets should that be something we wish to support in the future.

cpgserver/src/test/scala/io/shiftleft/cpgserver/cpg/DummyCpgProviderSpec.scala
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  "Creating a CPG" should {
+    "return a UUID referencing the eventual CPG" in withNewCpgProvider { cpgProvider =>
+      noException should be thrownBy cpgProvider.createCpg(Set.empty).unsafeRunSync()
+    }
+  }
+
+  "Retrieving a CPG" should {
+//    "return a success if the CPG was created successfully" in withNewCpgProvider { cpgProvider =>
```
fabsx00: These two tests are still commented out. Do they still not be implemented or should they be removed?

GlassAndOneHalf: Ah nice catch - I need to uncomment this.

cpgclientlib/tests/test.py
```diff
@@ -72,6 +72,7 @@ def testSimpleQuery(self):
         filename = os.path.join(SCRIPT_DIR, "testcode")
         self.client.create_cpg(filename)
         response = self.client.query("cpg.method.toJson")
+        print(response)
```
fabsx00: Debug message?

PR #371, opened 2019-10-01
Add Python as a valid language for CPG

ltcmelo:
Add Python as a valid language for CPG


hubertp:
@ltcmelo can you do the same here https://github.com/ShiftLeftSecurity/proto/blob/master/language/language.proto#L34 please?

ltcmelo:
Yes @hubertp , will do it now.

PR #392, opened 2019-10-10
Fix NPE

hubertp:


hubertp:
that doesn't spark confidence :)

PR #394, opened 2019-10-11
Bringing back changes 

hubertp:
The PR brings back changes that were reverted due to an urgent release and Sonatype publishing issues (the latter apparently still continuing).
The changes relate to "fast neighbor api".

This should now be safe to experiment with.


hubertp:
It was just an unfortunate timing, that's all.
Sonatype once managed to publish our artifact but I see that they are having issues. Again.
I will merge it now and, if needed, publish to our private repo.

PR #410, opened 2019-10-21
Move `location` support from codescience to codepropertygraph

fabsx00:
To allow https://github.com/ShiftLeftSecurity/joern/issues/118 to be implemented, this PR moves over location steps to `codepropertygraph`.


semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/package.scala
```diff
@@ -188,4 +187,22 @@ package object language {
   implicit def toCallForCallGraph(steps: Steps[nodes.Call]): Call =
     new Call(steps.raw)
 
+  // Locations
+
+  implicit def toExtendedNode(node: Node): ExtendedNode =
```
GlassAndOneHalf: I’m not sure if this was directly copied over but an `implicit class` would be a much better fit here.

PR #414, opened 2019-10-22
Added missing tests for `.location` step

fabsx00:
There were previously now tests for the `.location` step. This PR changes that.

### DIFF
--- a/semanticcpg/src/test/scala/io/shiftleft/semanticcpg/language/LocationTests.scala
+++ b/semanticcpg/src/test/scala/io/shiftleft/semanticcpg/language/LocationTests.scala
@@ -1,0 +1,90 @@
+package io.shiftleft.semanticcpg.language
+
+import org.scalatest.{Matchers, WordSpec}
+import io.shiftleft.semanticcpg.language._
+import io.shiftleft.semanticcpg.testfixtures.CodeToCpgFixture
+
+class LocationTests extends WordSpec with Matchers {
+
+  val code = """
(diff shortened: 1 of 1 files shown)


semanticcpg/src/test/scala/io/shiftleft/semanticcpg/language/LocationTests.scala
```diff
@@ -0,0 +1,90 @@
...
+      val locations = cpg.method.name("my_func").location.l
+      locations.size shouldBe 1
+
+      val loc = locations.head
+      loc.methodFullName shouldBe "my_func"
+      loc.methodShortName shouldBe "my_func"
+      loc.symbol shouldBe "my_func"
+      loc.lineNumber shouldBe Some(2)
+      loc.filename should endWith(".c")
+      loc.nodeLabel shouldBe "METHOD"
```
GlassAndOneHalf: If `loc` is a case class you can use `should matchPattern` here, e.g.

```scala
locations.head should matchPattern {
  case Location("my_func", "my_func", "my_func", Some(2), filename, "METHOD") if filename endsWith ".c" =>
}
```

fabsx00: Thanks for the tip! I played around with this a bit, but it becomes hard to see which fields we are checking. I agree that it's more compact, but ultimately, I would rather make this a bit more lengthy but easier to decipher.

PR #416, opened 2019-10-22
Lift CpgQueryExecutor into the console project.

GlassAndOneHalf:
This **does not** integrate the `CpgQueryExecutor` into the script manager, it simply moves the executor over to the `console` project.

### DIFF
--- a/console/build.sbt
+++ b/console/build.sbt
@@ -2,14 +2,71 @@ name := "console"
 
 enablePlugins(JavaAppPackaging)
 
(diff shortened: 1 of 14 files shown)


GlassAndOneHalf:
Max asked me to lift this component into the `console` project so that it can be used by both the server and the script manager, so it's mostly just moving code around. I won't merge until I've had a chat with him tomorrow to identify whether or not this is what we actually want to do.

PR #420, opened 2019-10-23
Adapt the script manager to CpgQueryExecutor

max-leuthaeuser:
The script manager now uses CpgQueryExecutor instead of its own implementation.

### DIFF
--- a/console/src/main/scala/io/shiftleft/console/query/DefaultCpgQueryExecutor.scala
+++ b/console/src/main/scala/io/shiftleft/console/query/DefaultCpgQueryExecutor.scala
@@ -17,10 +17,8 @@ import scala.concurrent.ExecutionContext
 class DefaultCpgQueryExecutor(scriptEngineManager: ScriptEngineManager)(implicit val cs: ContextShift[IO])
     extends CpgQueryExecutor[AnyRef] {
 
-  private val ENGINE_NAME = "scala"
-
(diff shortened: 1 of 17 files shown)


console/src/main/scala/io/shiftleft/console/query/DefaultCpgQueryExecutor.scala
```diff
@@ -2,27 +2,32 @@ package io.shiftleft.console.query
...
 class DefaultCpgQueryExecutor(scriptEngineManager: ScriptEngineManager)(implicit val cs: ContextShift[IO])
-    extends CpgQueryExecutor[String] {
+    extends CpgQueryExecutor[AnyRef] {
+
+  private val ENGINE_NAME = "scala"
 
-  private val engineType = "scala"
+  private lazy val engine: IO[ScriptEngine] = OptionT
+    .fromOption[IO](Option(scriptEngineManager.getEngineByName(ENGINE_NAME)))
+    .getOrElseF(IO.raiseError(new RuntimeException("Engine could not be instantiated.")))
```
GlassAndOneHalf: This can be a plain `val`. `IO` instances are only evaluated once they are `run`. We can probably inline the `ENGINE_NAME` on L20, too as it's only used here.

max-leuthaeuser: Fixed.

PR #421, opened 2019-10-23
Add missing test for AST traversals

fabsx00:
In the AST traversal tests, we were one test shy where we ensure that we can find multiplication and additions in call arguments. Added the test.


fabsx00:
Also added ability to traverse AST upwards only up to a certain root node via inAst and inAstMinusLeaf.

PR #425, opened 2019-10-23
allow to override _toJson easily

mpollmeier:


mpollmeier:
@GlassAndOneHalf fyi this failed on travis-ci until I increased the timeout for this (unrelated) test:  https://travis-ci.org/ShiftLeftSecurity/codepropertygraph/jobs/602064111#L3545

mpollmeier:
@GlassAndOneHalf I'd love to have less json frameworks in our dependencies and actually tried to get rid of Json4s, but couldn't agree with myself on a satisfying solution. Circe and upickle don't encode `Any`, but we currently encode by calling `node.valueMap`, which returns a map of properties of type `Map[String, Any]`, which json4s handles just fine. The efforts to ship a 'standard json ast' with scala haven't gone anywhere, and I shied away from making circe-core a dependency of codepropertygraph, simply because it's so low in the dependency tree. 

Anyway, on second thought maybe it's best if we do bring in circe-core as a dependency of codepropertygraph and generate the encoders as part of the DomainClassCreator. They'd be custom encoders because the nodes aren't case classes, but that's ok because it's generated code anyway. 
Thoughts?

mpollmeier:
@GlassAndOneHalf forgot to mention: our `.toJson` steps could then return the circe json ast (rather than strings like now). The repl could pretty print them nontheless.

PR #431, opened 2019-10-25
Cleanup NodeTypeStarters and add spec

fabsx00:
I am currently writing documentation on the query language and I'm taking this as an opportunity to add spec via tests and cleanup where required. In this PR, I clean up node type starters.
- atVerticesWithId is deprecated and replaced by `id`
- id accepts both a sequence of ids and a single id
- added `label` step to retrieve node label without using `raw`

### DIFF
--- a/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/NodeTypeStarters.scala
+++ b/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/NodeTypeStarters.scala
@@ -125,9 +125,18 @@ class NodeTypeStarters(cpg: Cpg) {
     new MethodRef(scalaGraph.V.hasLabel(NodeTypes.METHOD_REF).cast[nodes.MethodRef])
 
   /**
+  Begin traversal at node with id.
+    */
+  def id[NodeType <: nodes.StoredNode](anId: Any): NodeSteps[NodeType] = id(Seq(anId))
+
+  /**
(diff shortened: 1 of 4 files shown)


semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/NodeTypeStarters.scala
```diff
@@ -124,10 +124,19 @@ class NodeTypeStarters(cpg: Cpg) {
...
   /**
   Begin traversal at set of nodes - specified by their ids
     */
-  def atVerticesWithId[NodeType <: nodes.StoredNode](ids: Seq[Any]): NodeSteps[NodeType] =
-    if (ids.size == 0) new NodeSteps[NodeType](scalaGraph.V(-1).cast[NodeType])
+  def id[NodeType <: nodes.StoredNode](ids: Seq[Any]): NodeSteps[NodeType] =
+    if (ids.isEmpty) new NodeSteps[NodeType](scalaGraph.V(-1).cast[NodeType])
     else new NodeSteps[NodeType](scalaGraph.V(ids: _*).cast[NodeType])
+
+  @deprecated("October 2019", "")
```
GlassAndOneHalf: The date is second parameter of the `deprecated` annotation

fabsx00: thanks, fixed.

PR #434, opened 2019-10-25
Cleanup of tests for inclusion on website

fabsx00:
I would like to include some of our tests as-is on the website to help document the query language. In this PR, I therefore clean up the tests to make them shorter.

### DIFF
--- a/semanticcpg/src/test/scala/io/shiftleft/semanticcpg/language/LocationTests.scala
+++ b/semanticcpg/src/test/scala/io/shiftleft/semanticcpg/language/LocationTests.scala
@@ -1,7 +1,6 @@
 package io.shiftleft.semanticcpg.language
 
 import org.scalatest.{Matchers, WordSpec}
-import io.shiftleft.semanticcpg.language._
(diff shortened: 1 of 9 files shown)


semanticcpg/src/test/scala/io/shiftleft/semanticcpg/language/LocationTests.scala
```diff
@@ -11,8 +10,9 @@ class LocationTests extends WordSpec with Matchers {
       int x = foo(param1);
    }"""
 
-  "should return location for method" in {
-    CodeToCpgFixture().buildCpg(code) { cpg =>
+  CodeToCpgFixture(code) { cpg =>
```
GlassAndOneHalf: Nice! I guess this will give us some speedup?

fabsx00: yeah, I think so. It was a bit baffled to see that the CPG was calculated once per test. Don't see any downside to calculating it only once.

PR #437, opened 2019-10-28
Introduce `CodeDumper`

fabsx00:
This PR introduces a `.dump` step defined on all NodeSteps. This serves as a replacement for the old shell utility `joern-code`. It enables to quickly browse the code associated with findings. By integrating this capability into the query language, it becomes available both in Ocular and Joern without further integration work.

This fixes https://github.com/ShiftLeftSecurity/joern/issues/85

### DIFF
--- a/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/codedumper/CodeDumper.scala
+++ b/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/codedumper/CodeDumper.scala
@@ -52,7 +52,7 @@ object CodeDumper {
   /**
     * For a given `filename`, `startLine`, and `endLine`, return the corresponding code
     * by reading it from the file. If `lineToHighlight` is defined, then a line containing
-    * and arrow (as a source code comment) is included right before that line.
+    * an arrow (as a source code comment) is included right before that line.
     * */
   def code(filename: String, startLine: Integer, endLine: Integer, lineToHighlight: Option[Integer] = None): String = {
     val lines = Try(File(filename).lines.toList).getOrElse {
--- a/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/codedumper/CodeDumper.scala
+++ b/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/codedumper/CodeDumper.scala
@@ -1,0 +1,76 @@
+package io.shiftleft.semanticcpg.codedumper
+
+import io.shiftleft.codepropertygraph.generated.{NodeTypes, nodes}
+import io.shiftleft.semanticcpg.language._
+import io.shiftleft.semanticcpg.language.NodeSteps
+import better.files._
+import org.apache.logging.log4j.LogManager
+
+import scala.util.Try
+
+object CodeDumper {
+
+  private val logger = LogManager.getLogger(CodeDumper)
+  val arrow: CharSequence = "// ===>\n"
+
+  /**
+    * Evaluate the `step` and determine associated locations.
+    * Dump code at those locations
+    * */
+  def dump[NodeType <: nodes.StoredNode](step: NodeSteps[NodeType]): String =
+    step.location.l.map(dump).mkString("\n")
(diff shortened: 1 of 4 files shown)


semanticcpg/src/main/scala/io/shiftleft/semanticcpg/codedumper/CodeDumper.scala
```diff
@@ -0,0 +1,76 @@
...
+      .map { m =>
+        code(filename, m.lineNumber.get, m.lineNumberEnd.get, lineToHighlight)
+      }
+      .getOrElse("")
+  }
+
+  /**
+    * For a given `filename`, `startLine`, and `endLine`, return the corresponding code
+    * by reading it from the file. If `lineToHighlight` is defined, then a line containing
+    * and arrow (as a source code comment) is included right before that line.
```
Ferada:
```suggestion
    * an arrow (as a source code comment) is included right before that line.
```

fabsx00:
A short note on this feature: I have made this work for the Fuzzy C/C++ frontend so far, and a bit of work is required on each frontend that wants to support this feature.
- LINE_NUMBER and LINE_NUMBER_END need to be set correctly in methods
- LINE_NUMBER needs to be set correctly for expressions
- FILENAME in File nodes needs to be set correctly. I have used absolute paths in the C/C++ frontend. Do let me know if we should also support relative paths and a configurable code directory in which to search.

semanticcpg/src/test/scala/io/shiftleft/semanticcpg/codedumper/CodeDumperTests.scala
```diff
@@ -0,0 +1,46 @@
...
+  CodeToCpgFixture(code) { cpg =>
+    "should return empty string for empty traversal" in {
+      CodeDumper.dump(cpg.method.name("notinthere")) shouldBe ""
+    }
+
+    "should be able to dump complete function" in {
+      val query = cpg.method.name("my_func")
+      val code = CodeDumper.dump(query)
+      code should startWith(CodeDumper.arrow.toString)
+      code.contains("foo(param1)") shouldBe true
```
max-leuthaeuser: You may want to test with `code should include(someString)` or `code should include regex """someRegex"""` for increased readability of match results.

semanticcpg/src/test/scala/io/shiftleft/semanticcpg/codedumper/CodeDumperTests.scala
```diff
@@ -0,0 +1,46 @@
...
+      code should endWith("}")
+    }
+
+    "should dump method with arrow for expression (a call)" in {
+      val query = cpg.call.name("foo")
+      val code = CodeDumper.dump(query)
+      code should startWith("int")
+      code.matches(".*" + CodeDumper.arrow + ".*" + "int x = foo" + ".*")
```
max-leuthaeuser: See comment above.

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/codedumper/CodeDumper.scala
```diff
@@ -0,0 +1,76 @@
...
+
+    val lineToHighlight = location.lineNumber
+    method
+      .filter { m =>
+        m.lineNumber.isDefined && m.lineNumberEnd.isDefined
+      }
+      .map { m =>
+        code(filename, m.lineNumber.get, m.lineNumberEnd.get, lineToHighlight)
+      }
+      .getOrElse("")
```
max-leuthaeuser: Depending on the input size in comparison to the amount of data filtered, `collect` might be faster here as it does not create two intermediate lists. Other than that, its a matter of personal preference.
(see: https://stackoverflow.com/questions/36961777/scala-can-there-be-any-reason-to-prefer-filtermap-over-collect)

fabsx00:
Thanks, Max, all suggestions incorporated :)

PR #438, opened 2019-10-28
Syntax highlighting for `dump`.

fabsx00:
* Add syntax-highlighting using `source-highlight`
* Add checks to ensure that - for the time being - `dump` returns empty strings for all languages except C

### DIFF
--- a/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/codedumper/CodeDumper.scala
+++ b/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/codedumper/CodeDumper.scala
@@ -101,7 +101,7 @@ object CodeDumper {
       highlightedCode
     } catch {
       case exception: Exception =>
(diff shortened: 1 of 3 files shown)


semanticcpg/src/main/scala/io/shiftleft/semanticcpg/codedumper/CodeDumper.scala
```diff
@@ -71,4 +87,27 @@ object CodeDumper {
...
+    }
+
+    val f = File.newTemporaryFile("dump")
+    f.writeText(code)
+    val ret = try {
+      val highlightedCode: String = Process(Seq("source-highlight-esc.sh", f.path.toString, langFlag)).!!
+      highlightedCode
+    } catch {
+      case exception: Exception =>
+        logger.info("syntax highlighting now working. Is `source-highlight-esc.sh` installed?")
```
Ferada:
```suggestion
        logger.info("syntax highlighting now working. Is `source-highlight` installed?")
```

Issue #443, opened 2019-10-29
Mavencentral badge outdated


fabsx00:
The maven central badge shows that the latest version is 0.10.25, while really, the latest version is 0.10.96. Did this possibly break when we switched to publishing to `_2.12` directories?

https://maven-badges.herokuapp.com/maven-central/io.shiftleft/codepropertygraph/badge.svg

PR #445, opened 2019-10-29
fix maven badge

mpollmeier:
fixes https://github.com/ShiftLeftSecurity/codepropertygraph/issues/443


mpollmeier:
will hold back on merging due to sptests and wildebeest errors on master

Issue #441, opened 2019-10-28
`.dump` triggers exception in Ammonite


fabsx00:
`.dump` now supports syntax highlighting via `source-highlight`. Unfortunately, the escape sequences generated by `source-highlight` seem to not work together with Ammonite:

```
joern> cpg.method.name("malloc").callIn.dump 
java.lang.IllegalArgumentException: Unknown ansi-escape [00;38;05;70m at index 3 inside string cannot be parsed into an fansi.Str
  fansi.ErrorMode$Throw$.handle(Fansi.scala:419)
  fansi.ErrorMode$Throw$.handle(Fansi.scala:407)
  fansi.Str$.apply(Fansi.scala:272)
  fansi.Str$.implicitApply(Fansi.scala:227)
  pprint.Renderer.$anonfun$rec$27(Renderer.scala:136)
  pprint.Result$.fromString(Result.scala:53)
  pprint.Renderer.rec(Renderer.scala:136)
  pprint.PPrinter.tokenize(PPrinter.scala:110)
  ammonite.repl.FullReplAPI$Internal.print(FullReplAPI.scala:106)
  ammonite.repl.FullReplAPI$Internal.print$(FullReplAPI.scala:61)
  ammonite.repl.FullReplAPI$$anon$1.print(FullReplAPI.scala:34)
```

Note that the following works:
```
println(cpg.method.name("malloc").callIn.dump) 
``
as here, println interpretes the escape sequences.
Related post here: https://gitter.im/lihaoyi/Ammonite?at=5d1a93e19cbde24b2f59b509

This isn't a viable workaround for us though as we also want to make use of `browse`.

mpollmeier:
Update: I tried to fix this in different ways, but haven't found a satisfying solution yet:

1) make `.dump` return a different type (e.g. `Source` instead of `String`) and handling that in a separate pprint handler in BridgeBase.scala works, but unfortunately colouring happens afterwards
2) pprint uses `fansi` under the hood, which allows `fansi.ErrorMode.Strip` to skip any parsing errors. Unfortunately that also removes our already existent highlighting. 

I'm currently trying to construct an `fansi.Str` instance from the pre-parsed coloured string, and fixing up pprint to pass that through. 

So I haven't run out of ideas yet, but if you need a quick workaround, make `.dump` return `Unit` add a `println` inside that method. It's not really a solution though.

PR #447, opened 2019-10-31
fix display of already color-coded ansi in ammonite

mpollmeier:
We use source-highlight to encode source as ansi strings, e.g. the .dump step
Ammonite uses fansi for it's colour-coding, and while both pledge to follow the ansi codec, they aren't compatible

fixes https://github.com/ShiftLeftSecurity/codepropertygraph/issues/441

#### Testing with joern:
```
loadCpg("/home/mp/tmp/cpgtesting/linux-driver-small/cpg.bin.zip")
cpg.method.name("aac_src_ioremap").dumpc
browse(cpg.method.name("aac_src_ioremap").dumpc)
```


mpollmeier:
@fabsx00 please test with your initial example from your ticket - I think it's covered but better test before merging this. I guess that's from the linux kernel?

PR #453, opened 2019-11-01
size/count and is steps

mpollmeier:
Triggered https://github.com/ShiftLeftSecurity/codescience/issues/2995 it
became obvious that we're missing a few basic wrapper steps. These are easy
to add though.

`is(P)` is a bit of a leaky abstraction, but not too bad IMO.

n.b. independent of this we can still rename `filter` ->
`filterOnTrav` and `filterOnEnd` -> `filter`. Downside would be that we
have to teach customers. Upside is that we'll have to do that eventually
anyway, with the new collection-based traversal.


mpollmeier:
based on https://github.com/ShiftLeftSecurity/codepropertygraph/pull/452, review/merge that one first

fabsx00:
Closing for now in favor of extending nodes.$Foo types and renaming `filterOnEnd` to `where`.

PR #454, opened 2019-11-01
CodeDump: move arrow onto the right side

fabsx00:
* Moving arrow onto the right-hand side so that code formatting isn't broken.
* Update fuzzyc version to latest for line-number patch.


fabsx00:
This build will become green once sonatype allows publishing of artifacts again:
https://status.maven.org/incidents/xgnvrnjmn22v?u=9tb7bh9vxczx

PR #459, opened 2019-11-04
Introduce `where` as replacement for `filterOnEnd`

fabsx00:
Found a non-intrusive solution to https://github.com/ShiftLeftSecurity/codescience/issues/2995: renamed `filterOnEnd` to `where`. I left `filterOnEnd` as an alias for now and deprecated it.


ml86:
Not change the `filter` signature is rather a nice way of not running into the backward compat problem in the first place. ;)
I could live with it but the term `where` always lets me think about SQL and gives me the creeps.

mpollmeier:
I'm okay with this, just note that we're only delaying the change, since the new dsl will have `filter(NodeType => Boolean)`, i.e. basically an alias for `where`. 

I'll create a separate PR for the size/count steps from the original proposal https://github.com/ShiftLeftSecurity/codepropertygraph/pull/453

PR #461, opened 2019-11-06
Add ARGUMENT edge.

ml86:
The ARGUMENT edge are between CALL nodes and their argument and
between RETURN nodes and the return expression.

Formerly the AST edges indicated the "argument" relation but this
became unfeasable through the introduction of call receivers which
are not arguments to a call but still need to be ast children of it.

A backward compat pass is also in this PR.


fabsx00:
If we traverse AST edges, can we still reach arguments by expanding on the CALL node? This would be a requirement for our AST query language to work.

ml86:
Yes, but you might get more than the arguments. For the C call: funcPtr(a,b) you would get "funcPtr, a, b" as AST children and "a, b" when expanding via ARGUMENT edge.

dataflowengine/src/main/scala/io/shiftleft/dataflowengine/language/TrackingPoint.scala
```diff
@@ -94,12 +94,12 @@ class TrackingPoint(raw: GremlinScala[nodes.TrackingPoint]) extends NodeSteps[no
   private def getTrackingPoint(vertex: Vertex): Option[nodes.TrackingPoint] = {
     vertex match {
       case identifier: nodes.Identifier =>
-        getTrackingPoint(identifier._astIn.nextChecked)
+        getTrackingPoint(identifier._argumentIn().nextChecked)
```
fabsx00: I recently noticed these new _foo methods. Where are they coming from? When did we introduce them? I'm not a fan of this, we should rather define these kinds of methods as part of the query language.

fabsx00:
> Yes, but you might get more than the arguments. For the C call: funcPtr(a,b) you would get "funcPtr, a, b" as AST children and "a, b" when expanding via ARGUMENT edge.

Right, but that's the same behavior we had before this PR, right?

dataflowengine/src/main/scala/io/shiftleft/dataflowengine/language/TrackingPoint.scala
ml86: Bernhard implemented those a few weeks back. They have been introduced as replacement for the low level gremlin API `vertices(<direction>, <edgeTypes>)` in order to increase performance.

ml86:
> > Yes, but you might get more than the arguments. For the C call: funcPtr(a,b) you would get "funcPtr, a, b" as AST children and "a, b" when expanding via ARGUMENT edge.
> 
> Right, but that's the same behavior we had before this PR, right?

Yes, but in the past ONLY arguments have been the ast children of CALL nodes which wont be true anymore in the future.

fabsx00:
That's fine.

PR #464, opened 2019-11-06
move to bintray

mpollmeier:


mpollmeier:
fyi there's an ongoing support ticket on bintray re fuzzyc2cpg's inclusion in jcenter. I'll only merge this once we're all clear on that front

PR #465, opened 2019-11-07
Dominator/CDG pass

fabsx00:
The domanitor and CDG passes were for some reason in `semanticcpg-ext`, although these are passes required for a PDG - which is a structure present in the original 2014 code property graph. Moving the code over.

### DIFF
--- a/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/passes/cfgdominator/CfgDominator.scala
+++ b/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/passes/cfgdominator/CfgDominator.scala
@@ -1,0 +1,114 @@
+package io.shiftleft.semanticcpg.passes.cfgdominator
+
+class CfgDominator[Node](adapter: CfgAdapter[Node]) {
+
+  /**
+    * Calculates the immediate dominators of all CFG nodes reachable from cfgEntry.
+    * Since the cfgEntry does not have an immediate dominator, it has no entry in the
+    * return map.
+    *
+    * The used algorithm is from: "A Simple, Fast Dominance Algorithm" from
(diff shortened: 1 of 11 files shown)


semanticcpg/src/main/scala/io/shiftleft/semanticcpg/passes/cfgdominator/CfgDominator.scala
```diff
@@ -0,0 +1,114 @@
...
+        }
+
+        val nodeIndex = indexOf(node)
+        if (dominators(nodeIndex) != newImmediateDominator) {
+          dominators(nodeIndex) = newImmediateDominator
+          changed = true
+        }
+      }
+
+    }
```
GlassAndOneHalf: For the future, we can eliminate the `var` here by replacing:
- L38 with `while (true)` 
- L59 with `continue`
- L62 with `break`

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/passes/cfgdominator/DomTreeAdapter.scala
```diff
@@ -0,0 +1,13 @@
+package io.shiftleft.semanticcpg.passes.cfgdominator
+
+trait DomTreeAdapter[Node] {
+
+  /**
+    * Returns the immediate dominator of a cfgNode. The returned value
+    * can be None if cfgNode was the cfg entry node while calculating
+    * the dominator relation or if cfgNode is dead code. In the post
+    * dominator case "dead code" means code which does lead to the
```
GlassAndOneHalf:
```suggestion
    * dominator case "dead code" means code which does not lead to a
```

PR #469, opened 2019-11-11
Queries for complexity metrics

fabsx00:
I am bringing in examples of queries, and I'm making them unit tests so they continue to work. This PR brings in queries for complexity metrics.

### DIFF
--- a/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/nodemethods/AstNodeMethods.scala
+++ b/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/nodemethods/AstNodeMethods.scala
@@ -18,4 +18,19 @@ class AstNodeMethods(val node: nodes.AstNode) extends AnyVal {
     * */
   def astMinusRoot: AstNode = node.start.astMinusRoot
 
+  def isControlStructure : Boolean = node.start.isControlStructure.size == 1
+
+  def depth(p : nodes.AstNode => Boolean = _ => true) : Int = {
+
+    val additionalDepth = if (p(node)) { 1 } else { 0 }
+
+    val childDepths = node.start.astChildren.map(_.depth(p)).l
+
+    additionalDepth + (if (childDepths.isEmpty) {
+      0
+    } else {
+      childDepths.max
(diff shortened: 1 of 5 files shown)


semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/nodemethods/AstNodeMethods.scala
```diff
@@ -18,4 +18,19 @@ class AstNodeMethods(val node: nodes.AstNode) extends AnyVal {
     * */
   def astMinusRoot: AstNode = node.start.astMinusRoot
 
+  def isControlStructure: Boolean = node.start.isControlStructure.size == 1
```
Ferada: I, what?

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/nodemethods/AstNodeMethods.scala
```diff
@@ -18,4 +18,19 @@ class AstNodeMethods(val node: nodes.AstNode) extends AnyVal {
...
+
+  def depth(p: nodes.AstNode => Boolean = _ => true): Int = {
```
Ferada: Also these should be commented I guess.

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/nodemethods/AstNodeMethods.scala
```diff
@@ -18,4 +18,19 @@ class AstNodeMethods(val node: nodes.AstNode) extends AnyVal {
...
+
```
Ferada:
```suggestion
    node.start.astChildren.map(_.depth(p)).l.fold(if (p(node)) { 1 } else { 0 })(Math.max)
```
At least the last line can be a fold instead.

fabsx00: As discussed offline, leaving it as-is for now.

PR #480, opened 2019-11-14
Cleanup of sample scripts and operator extension

fabsx00:
- After playing with this for a bit, I decided to not "extend" call nodes via operator nodes for now. Meanwhile, @mpollmeier has fixed the ODB bug, which lead to us using implicit conversions for this purpose.
- Cleaned up the remainder of the operator classes I brought in yesterday to work like `ArrayAccess`
- Adapted sample queries


semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/operatorextension/ArithmeticTrav.scala
```diff
@@ -0,0 +1,6 @@
+package io.shiftleft.semanticcpg.language.operatorextension
+
+import gremlin.scala.GremlinScala
+import io.shiftleft.semanticcpg.language.Steps
+
+class ArithmeticTrav(raw: GremlinScala[nodes.Arithmetic]) extends Steps[nodes.Arithmetic](raw) {}
```
GlassAndOneHalf: Can this class be deleted?

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/operatorextension/nodes/Arithmetic.scala
```diff
@@ -0,0 +1,5 @@
+package io.shiftleft.semanticcpg.language.operatorextension.nodes
+
+import io.shiftleft.codepropertygraph.generated.nodes
+
+class Arithmetic(val call: nodes.Call) extends AnyRef {}
```
GlassAndOneHalf: Ditto - is there any reason to have an empty class? The `extends AnyRef` can also be dropped as it is implicit.

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/operatorextension/nodes/ArrayAccess.scala
```diff
@@ -4,12 +4,10 @@ import io.shiftleft.codepropertygraph.generated.nodes
 import io.shiftleft.semanticcpg.language._
 import io.shiftleft.semanticcpg.language.types.expressions.Identifier
 
-class ArrayAccess(node: nodes.Call) {
+class ArrayAccess(val call: nodes.Call) extends AnyRef {
```
GlassAndOneHalf: Ditto w/ the `AnyRef`. There are a few others below, too.

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/package.scala
```diff
@@ -219,13 +219,12 @@ package object language {
   implicit def toNodeTypeStartersOps(cpg: Cpg): OpNodeTypeStarters =
     new OpNodeTypeStarters(cpg)
 
-  implicit def arrayAccessToCall(
-      node: operatorextension.nodes.ArrayAccess): io.shiftleft.semanticcpg.language.nodemethods.CallMethods =
-    node.call
+  implicit def toOpsAstNodeTrav2[NodeType <: nodes.AstNode](
```
GlassAndOneHalf: Might be worth dropping the `2` in the name, I'm not too sure what it means here!

fabsx00:
Concerning empty classes: they are not entirely empty, they do have that `call` field. The reason I introduce them now already is that it is inconsistent to have some of the classes wrap a type from `nodes`, while others bring in their own node. Also, methods in these classes will follow very soon. I can hold off on merging this one though until these methods are present.

PR #491, opened 2019-11-21
New pass for calculating IS_METHOD_NEVER_OVERRIDDEN flag

lifecoder:


semanticcpg/src/main/scala/io/shiftleft/semanticcpg/passes/BindingMethodOverridesPass.scala
```diff
@@ -0,0 +1,39 @@
...
+import io.shiftleft.passes.{CpgPass, DiffGraph}
+import io.shiftleft.semanticcpg.language._
+import org.apache.logging.log4j.LogManager
+
+class BindingMethodOverridesPass(cpg: Cpg) extends CpgPass(cpg) {
+  override def run(): Iterator[DiffGraph] = {
+    cpg.method.map(processMethod).toList().iterator
+  }
+
+  def processMethod(m: nodes.Method): DiffGraph = {
```
ml86: No single letter abbreviations please. `method` seems to be a more appropriate name.

lifecoder: Done

PR #492, opened 2019-11-21
WIP: enforce exhaustive pattern matching

mpollmeier:


mpollmeier:
@ml86 this breaks codescience' `FreeTrackingPoint`, which extends `nodes.TrackingPoint`
It seems to be an implementation detail of the dataflow algorithm, so we probably don't want to have it in the cpg schema, right? Can we do that implementation differently? We'll only get exhaustive matching checks with sealed traits.

ml86:
If we need sealed traits this is a problem. I would need to create an extra wrapper class around all nodes.TrackingPoint in order to get similar behaviour. This is not only "not nice" in terms of implementation but also would cost us quite a lot of memory.
The other option is defining FreeTrackingPoint in enhancement.json and than us the generated class instead of the current FreeTrackingPoint. I guess this would work but i will need to test it.

mpollmeier:
We can only enforce this for sealed traits, at least with scalac (which sbt uses). intellij may behave differently since they reimplemented scalac. 

The extra wrapper doesn't sound like a good idea. Adding `FreeTrackingPoint` to the schema sounds good to me, you'd be best to judge if it's worth it. 

In the meantime I'll extract the changes unrelated to the `sealed trait` issue into a separate PR.

mpollmeier:
The uncontroversial changes are in master and I rebased this PR to just adding `sealed`.

fabsx00:
Nothing has happened on this one since November. Closing. Reopen if needed.

PR #497, opened 2019-11-24
Improvements for operator extension + truncation example

fabsx00:
A sample query for identifying truncations in `strlen` assignments, and a few improvements for the operator extension.


semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/operatorextension/nodes/AssignSource.scala
```diff
@@ -0,0 +1,5 @@
+package io.shiftleft.semanticcpg.language.operatorextension.nodes
+
+import io.shiftleft.codepropertygraph.generated.{nodes => basenodes}
+
+class AssignSource(val expr: basenodes.Expression) extends AnyRef {}
```
GlassAndOneHalf:
```suggestion
class AssignSource(val expr: basenodes.Expression) {}
```

PR #499, opened 2019-11-25
Add referencedMethod step to METHOD_REF.

ml86:


semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/types/expressions/MethodRef.scala
```diff
@@ -1,12 +1,20 @@
...
 class MethodRef(raw: GremlinScala[nodes.MethodRef])
     extends NodeSteps[nodes.MethodRef](raw)
     with ExpressionBase[nodes.MethodRef]
-    with LineNumberAccessors[nodes.MethodRef] {}
+    with LineNumberAccessors[nodes.MethodRef] {
+
+  /**
+    * Traverse to referenced method.
+    * */
+  def referencedMethod: Method =
```
fabsx00: Can we rename this to `method`?

ml86: Sure. Done.

ml86: Actually we cannot. This is because MethodRef is inheriting from ExpressionBase which provides a method function. Thats why i suggest we stick with referencedMethod.

PR #505, opened 2019-11-28
New diffgraph and parallel execution

lifecoder:
Changes:
* There are lots of empty or single element `DiffGraph`, so new implementation
has separate instances for empty or single element diffs which should take less memory.
* DiffGraph is split into DiffGraph.Builder which resembles old API and immutable diffgraph. This is done to add `+` operation to combine immutable diffgraphs later on and to avoid allocating node/edge/properties buffers for each diffgraph (lots of passes add just properties or just edges).


codepropertygraph/src/main/scala/io/shiftleft/passes/CpgPass.scala
```diff
@@ -39,6 +42,7 @@ import io.shiftleft.codepropertygraph.Cpg
   */
 abstract class CpgPass(cpg: Cpg) {
   import CpgPass.logger
+  implicit val executionContext: ExecutionContext = scala.concurrent.ExecutionContext.Implicits.global
```
mpollmeier: (minor) i think `import scala.concurrent.ExecutionContext.Implicits.global` does the same, no need for a local implicit val

codepropertygraph/src/main/scala/io/shiftleft/passes/CpgPass.scala
```diff
@@ -148,74 +152,77 @@ private[passes] case class IdentityHashWrapper[T <: AnyRef](value: T) {
...
+      .newBuilder() // TODO: Can we cache builders???
+      .setKey(nodeId)
+      .setType(NodeType.valueOf(node.label))
 
-      val finalNode = nodeBuilder.build()
-      builder.addNode(finalNode)
-    }
+    node.properties
+      .foreach {
+        case (key, value) if !key.startsWith("_") =>
```
mpollmeier: did you test this somewhere? this will throw a MatchError for all other cases (`if key.startsWith("_")`)
demo:
```
List(1,2,3).foreach { case i if i < 3 => println(i) }
1
2
scala.MatchError: 3 (of class java.lang.Integer)
```

codepropertygraph/src/main/scala/io/shiftleft/passes/ParallelIteratorExecutor.scala
```diff
@@ -1,11 +1,13 @@
...
 
+import scala.collection.mutable
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.duration.Duration
 import scala.concurrent.{Await, Future}
+import scala.reflect.ClassTag
 
-class ParallelIteratorExecutor[T](iterator: Iterator[T]) {
-  def map[D](func: T => D): Iterator[D] = {
+class ParallelIteratorExecutor[T: ClassTag](iterator: Iterator[T]) {
```
mpollmeier: `ClassTag` doesn't seem to be needed here?

codepropertygraph/src/main/scala/io/shiftleft/passes/CpgPass.scala
lifecoder: This is the way it is now, I've just moved this code from few lines above.

codepropertygraph/src/main/scala/io/shiftleft/passes/CpgPass.scala
lifecoder: Yes, I've removed it

codepropertygraph/src/main/scala/io/shiftleft/passes/ParallelIteratorExecutor.scala
lifecoder: Correct, I've removed it

codepropertygraph/src/main/scala/io/shiftleft/passes/CpgPass.scala
mpollmeier: ok, so we expect this to crash and burn when the property name starts with `_`? sounds a bit odd.
i guess it's ok because no diffgraph would/should ever create such a property, but i'm not sure why we'd do this check in the first place. should we better just remove the `if !key.startsWith("_")` condition @fabsx00 ?

PR #510, opened 2019-12-06
Ignoring UNKNOWN nodes while validating facts and keys.

max-leuthaeuser:


cpgvalidator/src/main/scala/io/shiftleft/cpgvalidator/validators/InFactsValidator.scala
```diff
@@ -24,27 +24,28 @@ class InFactsValidator(errorRegistry: ValidationErrorRegistry) extends Validator
...
-            val dstNode = traverser.get
-            val inEdges = dstNode.edges(Direction.IN).asScala.toList
-            inFactsByEdgeType.foreach {
-              case (edgeType, inFacts) =>
-                val actualSrcNodes =
-                  inEdges.filter(_.label == edgeType).map(_.outVertex)
-                inFacts.foreach { inFact =>
-                  validateInDegree(dstNode, actualSrcNodes, inFact)
+            traverser.get match {
+              case dstNode if dstNode.label() != "UNKNOWN" =>
```
ml86: Please use NodeTypes.UNKNOWN instead of "UNKNOWN".

ml86: Also fix other places where this happens.

PR #516, opened 2019-12-10
replace nodevisitor pattern by pattern matches

mpollmeier:
* jmh tests revealed that pattern matching is actually faster than the
visitor pattern, mostly because the visitor pattern results in a
virtual method call, which can't be inlined (at least not for >2 cases)
* cleaner and more condense code
* original PR: https://github.com/ShiftLeftSecurity/codepropertygraph/pull/317


semanticcpg/src/main/scala/io/shiftleft/semanticcpg/utils/ExpandTo.scala
```diff
@@ -61,8 +61,12 @@ object ExpandTo {
   }
 
   def methodToFormalReturn(method: Vertex): nodes.StoredNode = {
+    println(method.getClass)
+    println(method.isInstanceOf[nodes.StoredNode])
     method
       .asInstanceOf[nodes.StoredNode]
+//      .asInstanceOf[nodes.Method]
+//      .get
```
ml86: Remove debug code.

mpollmeier: you passed the test :)

thanks, done!

PR #518, opened 2019-12-11
Fixes for the CPG validator

max-leuthaeuser:
Fixes for the CPG validator:
 * do not ignore NodeTypes.UNKNOWN as src/dst nodes during in/out facts validation any more
 * added NodeTypes.UNKNOWN to possible in/out nodes for ARGUMENT edges

### DIFF
--- a/cpgvalidator/src/main/scala/io/shiftleft/cpgvalidator/validators/InFactsValidator.scala
+++ b/cpgvalidator/src/main/scala/io/shiftleft/cpgvalidator/validators/InFactsValidator.scala
@@ -31,7 +31,7 @@ class InFactsValidator(errorRegistry: ValidationErrorRegistry) extends Validator
                 inFactsByEdgeType.foreach {
                   case (edgeType, inFacts) =>
                     val actualSrcNodes =
-                      inEdges.filter(_.label == edgeType).map(_.outVertex).filterNot(_.label() == NodeTypes.UNKNOWN)
+                      inEdges.filter(_.label == edgeType).map(_.outVertex)
                     inFacts.foreach { inFact =>
                       validateInDegree(dstNode, actualSrcNodes, inFact)
                     }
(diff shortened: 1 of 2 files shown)


cpgvalidator/src/main/scala/io/shiftleft/cpgvalidator/facts/InFactsImporter.scala
```diff
@@ -23,28 +23,28 @@ class InFactsImporter extends FactsImporter {
       NodeTypes.LITERAL has 1 incoming EdgeTypes.AST from SuperTypes.Expression or NodeTypes.CONTROL_STRUCTURE,
       NodeTypes.LITERAL has 0 to N incoming EdgeTypes.CFG from SuperTypes.Expression or NodeTypes.METHOD,
       NodeTypes.LITERAL has 0 to 1 incoming EdgeTypes.RECEIVER from NodeTypes.CALL,
-      NodeTypes.LITERAL has 0 to 1 incoming EdgeTypes.ARGUMENT from NodeTypes.CALL or NodeTypes.RETURN,
+      NodeTypes.LITERAL has 0 to 1 incoming EdgeTypes.ARGUMENT from NodeTypes.CALL or NodeTypes.RETURN or NodeTypes.UNKNOWN,
```
ml86: Why do we need to put UNKNOWN here if we special case it in the checker?

max-leuthaeuser: I removed that special filtering in the checker. Otherwise, UNKNOWN nodes would not  be taken into account when checking edges.

ml86: You are right the filtering at this point is wrong. But please have a look at the `validateAllDstNodeTypes` method. There we want to also allow UNKNOWN nodes.

max-leuthaeuser: True! Fixed that with the latest commits here.
With that, no changes to the facts are needed.

PR #520, opened 2019-12-14
Shorthands for NodeTypeStarters

fabsx00:
Some time in October, we brought in `cpg.call($name)` as a shorthand for `cpg.call.name($name)`. While writing queries, I have been using this quite extensively, and soon found that similar shorthands would make sense for almost all NodeTypeStarters. Also, there is usually one field that is intuitively the primary field one would filter on, e.g., `name` for `call`, `fullName` for `method` or `name` for member.

This PR introduces shorthands for NodeTypeStarters. Roughly, the rule is: if there is a `fullName` field, filter on that, if not filter on `name`, and if that doesn't exist, filter on `code`. An exception is `methodRef`, where we filter on the fullName of the referenced method.

Minor edit: renamed `returnExpression` to `returns`. The correct name would be `return`, however, `return` is a reserved keyword. The rule we used in the case where the correct function name is a reserved keyword was to switch to plural, e.g., `types` as opposed to `type`. Consequently, it must be `returns`, not `returnExpression`.

### DIFF
--- a/console/src/test/scala/io/shiftleft/console/ScriptManagerTest.scala
+++ b/console/src/test/scala/io/shiftleft/console/ScriptManagerTest.scala
@@ -39,20 +39,21 @@ class ScriptManagerTest extends WordSpec with Matchers with Inside {
       val scripts = scriptManager.scripts()
       val expected = List(
         ScriptCollections("general",
-          ScriptDescriptions(
-            "A collection of general purpose scripts.",
-            List(ScriptDescription("list-funcs", "Lists all functions."))
-          )),
+                          ScriptDescriptions(
+                            "A collection of general purpose scripts.",
+                            List(ScriptDescription("list-funcs", "Lists all functions."))
+                          )),
         ScriptCollections("java",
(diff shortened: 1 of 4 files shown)


fabsx00:
Concerning regex: with our current query language, we've gone the route that all strings for property filters are considered to be regular expressions except for the `$fieldExact` accessors, e.g., in `cpg.method.fullNameExact("my.package.foo")`, we match only if `fullName` is equal to `my.package.foo`. 

I believe the reason for going this direction in the first place was that the regex case is the more common of the two cases, in particular for `fullName`, as `fullName` is pretty lengthy. So, we wanted queries that use regex to look a bit more elegant than the ones where strings are used. The additional query language features proposed here follow this generic idea of making regex the default. I would propose we leave it like this, but I'm open for opinions.

mpollmeier:
Totally agree with all of the above - in hindsight I would prefer we had gone the other way around, as @AlexDenisov suggested. But making a change like that now will confuse a lot of people, so it's probably not a good idea :(

PR #525, opened 2019-12-19
Ignore missing method binding nodes when checking for overridden methods.

lifecoder:


semanticcpg/src/main/scala/io/shiftleft/semanticcpg/passes/BindingMethodOverridesPass.scala
```diff
@@ -17,14 +19,18 @@ class BindingMethodOverridesPass(cpg: Cpg) extends CpgPass(cpg) {
...
+          val neverOverriddenFlag = isMethodNeverOverridden(typeDecl.get, method.name, method.signature)
+          diff.addNodeProperty(
+            binding,
+            NodeKeys.IS_METHOD_NEVER_OVERRIDDEN.name,
+            neverOverriddenFlag.asInstanceOf[AnyRef]
+          )
+        } else {
+          logger.error(
+            "No binding typeDecl found in BindingMethodOverridesPass: " + method.name + " " + method.signature)
+          true
```
Ferada: With `foreach` don't need the return value do you?

lifecoder: @Ferada Correct, thanks

PR #530, opened 2020-01-06
Allow for multiple `BINDS`.

Ferada:
Makes it so that not all of the bindings need to have a method with it, instead the first one with a matching method will be returned instead, which should be backwards-compatible with existing CPGs, but now also allows us to find flows from `go2cpg` for simple method references, c.f. <https://github.com/ShiftLeftSecurity/codescience/pull/3230> for the now working example.


Ferada:
As discussed, this shouldn't actually be necessary and isn't after a few changes in `go2cpg`.

PR #532, opened 2020-01-07
disintegrate ExpandTo: AstNode.toCallOrReturn

mpollmeier:


semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/nodemethods/AstNodeMethods.scala
```diff
@@ -55,4 +58,17 @@ class AstNodeMethods(val node: nodes.AstNode) extends AnyVal {
                        })
   }
 
+  def toCallOrReturn: nodes.Expression = _toCallOrReturn(node)
```
fabsx00: This method would then be available on all AST nodes, right? The convention is that we drop the `to` if we traverse to another node, and use `as` as a prefix, if this is a cast. This looks like a case where we traverse to a node. It's unclear from the name though how the call or return is related to the AST node. I take it it's the call that wraps this?

mpollmeier: Yes we do traverse here, so `as` would be the wrong prefix.
I only refactored the code, but essentially `_toCallOrReturn` traverses to it's parent expression (e.g. call or return) by following the incoming AST nodes. It's continuing it's walk until it hits an expression that's not a generic "member access operation", e.g., "<operator>.memberAccess".

If you prefer a more generic name, maybe `toParentExpression` would work, but @ml86 may have a better idea. Just let me know what you would like.

ml86:
@mpollmeier we cannot disintegrate ExpandTo in that way. I introduced it for a reason which was performance and now i fear that we might run into performance problems if the change is in a hot path.

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/nodemethods/AstNodeMethods.scala
fabsx00: `parentExpression` then, without the `to`. This is consistent with e.g., `method.parameter` as opposed to `method.toParameter`.

mpollmeier: done

mpollmeier:
@ml86 I measured the invocation count of this specific method (`AstNodeMethods.parentExpression`, formerly `ExpandTo.argumentToCallOrReturn`) that this PR removes: cpg2sp on servlettarpit counted 866 invocations. That doesn't sound like it's on the hot path...

Given that it's now in an extension method via a value class, it shouldn't incur any additional garbage objects, but I haven't measured that (wouldn't really matter for that small number). 

How about we go ahead with this, and I'll do the same invocation counts for the other methods before I remove them?

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/nodemethods/AstNodeMethods.scala
```diff
@@ -55,4 +58,17 @@ class AstNodeMethods(val node: nodes.AstNode) extends AnyVal {
                        })
   }
 
+  def parentExpression: nodes.Expression = _parentExpression(node)
```
ml86: There is an `astParent` method in AstNode.scala. But no corresponding version of it on AstNodeMethods.scala. I guess that is why you did not the the very similar naming.
Please add a comment what `parentExpression` does and how is compares to `astParent`.
Please also add a version of `parentExpression` to AstNode.scala in order to no let those to diverged even further.

mpollmeier: done in 8f5090e8ff5d4ea44a9db445c26bfd7135f043e0

fabsx00:
@GlassAndOneHalf this seems relevant for your work on AST traversals as well.

PR #537, opened 2020-01-08
Operator extension cleanup

GlassAndOneHalf:
- Remove unused `Arithmetic` and `Assignment` types.
- Remove all node wrapper types, replacing them with implicit classes.
- Rename `Target` to `TargetTrav` for consistency.


queries/src/test/scala/io/shiftleft/queries/MallocMemcpyTests.scala
```diff
@@ -30,7 +30,7 @@ class MallocMemcpyTests extends WordSpec with Matchers {
       * */
     "find calls to malloc/memcpy system with different expressions in arguments" in {
 
-      val src = cpg.call("malloc").filter(_.argument(1).arithmetics).l
+      val src = cpg.call("malloc").filter(_.argument(1)).l
```
fabsx00: I'm not sure about this one: wasn't the point of this query to look for arithmetic operations in first arguments to malloc?

GlassAndOneHalf: Added a commit adding that back.

PR #541, opened 2020-01-12
Diffgraph refactoring

lifecoder:


lifecoder:
@mpollmeier Fixed the issue that broke sptests last time I've merged this.
The cause was that before we ignored when the same node is added more than once.
It happens in various places (which could be fixed), for example like this:
```
  private def addSensitiveVariables(sensitiveData: SensitiveData)(
      implicit dstGraph: DiffGraph.Builder) = {
      ...
      dstGraph.addNode(node) // stores node in diffgraph
      node.start.store // stores it again and also its children
```

mpollmeier:
this time around i let you do the merge :)

PR #542, opened 2020-01-13
Add replacement for default parameter for tagging.

Ferada:


fabsx00:
Good idea. Makes it less clumsy.

PR #548, opened 2020-01-17
provide steps via implicit conversions rather than inheritance

mpollmeier:
Going forward (and moving towards overflowdb-traversal) I think it
would simplify matters if we used a single concept to provide
node-specific steps.
The most flexible and commonly used (by us) mechanism is extension via implicit conversion.

To test the waters, this PR only tackles one simple case. Once approved,
I'll roll this out to all PropertyAccessors as well as all base trait
steps (e.g. `generalizations.Expression`).

### DIFF
--- a/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/package.scala
+++ b/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/package.scala
@@ -3,7 +3,7 @@ package io.shiftleft.semanticcpg
 import gremlin.scala._
 import io.shiftleft.codepropertygraph.Cpg
 import io.shiftleft.codepropertygraph.generated.nodes
-import io.shiftleft.codepropertygraph.generated.nodes.{HasFullName, HasLineNumber, HasLineNumberEnd, HasName, Node, StoredNode}
+import io.shiftleft.codepropertygraph.generated.nodes.{HasFullName, HasIsExternal, HasLineNumber, HasLineNumberEnd, HasName, Node, StoredNode}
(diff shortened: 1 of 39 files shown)


semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/package.scala
```diff
@@ -132,6 +132,10 @@ package object language {
   implicit def toComment(steps: Steps[nodes.Comment]): Comment =
     new Comment(steps.raw)
 
+  implicit def toLineNumberEndAccessors[A <: StoredNode with HasLineNumberEnd](
```
GlassAndOneHalf: Is there any reason we choose implicit conversions over implicit classes?

ml86: Dont know whether this was Micheal motivation but using the implicit methods allows you to split the class definition from the implicit conversion functionality which lets you group the implicit stuff in common places like we do with `io.shiftleft.semanticcpg.language._`.

mpollmeier: Yes exactly, the main idea is to have the implicit conversions in few places, so that in ocular/joern we just use the `language._` import and have everything at the user's fingertips. That pattern is used across all of our modules. 

That's not to say that implicit classes are bad, but it's already quite a burden to keep our downstream projects in sync, and with the implicits scattered in more places we'd have a harder time maintaining ocular/joern imports. Makes sense?

PR #550, opened 2020-01-19
Allow application of enhancements without serialization

fabsx00:
Prior to this PR, we called the EnhancementsRunner and DataFlowRunner in a few places where we needed enhancements/dataflow to be applied to the CPG, but we did not want them to be serialized into proto. We achieved this by simply passing a SerializedCpg without a filename. While this works, it means that the enhancementsRunner still serializes, it just does not write the serialized files anywhere. As serialization comes at a cost, with this PR, we now allow passing in no SerializedCpg at all, in which case the EnhancementRunner/DataflowRunner only applies but does not serialize the graph.

This change is backward compatible.

### DIFF
--- a/dataflowengine/src/main/scala/io/shiftleft/dataflowengine/layers/dataflows/DataFlowRunner.scala
+++ b/dataflowengine/src/main/scala/io/shiftleft/dataflowengine/layers/dataflows/DataFlowRunner.scala
@@ -8,9 +8,16 @@ import io.shiftleft.dataflowengine.semanticsloader.Semantics
 
 class DataFlowRunner(semantics: Semantics) {
 
-  def run(cpg: Cpg, serializedCpg: SerializedCpg): Unit = {
+  def run(cpg: Cpg, serializedCpg: SerializedCpg): Unit = run(cpg, Some(serializedCpg))
+
+  def run(cpg: Cpg, serializedCpg: Option[SerializedCpg] = None): Unit = {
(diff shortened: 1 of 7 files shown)


dataflowengine/src/main/scala/io/shiftleft/dataflowengine/layers/dataflows/DataFlowRunner.scala
```diff
@@ -8,9 +8,16 @@ import io.shiftleft.dataflowengine.semanticsloader.Semantics
...
 class DataFlowRunner(semantics: Semantics) {
 
-  def run(cpg: Cpg, serializedCpg: SerializedCpg): Unit = {
+  def run(cpg: Cpg, serializedCpg: SerializedCpg): Unit = run(cpg, Some(serializedCpg))
+
+  def run(cpg: Cpg, serializedCpg: Option[SerializedCpg] = None): Unit = {
     val enhancementExecList = List(new PropagateEdgePass(cpg, semantics), new ReachingDefPass(cpg))
-    enhancementExecList.foreach(_.createApplySerializeAndStore(serializedCpg))
+
+    if (serializedCpg.isDefined) {
```
mpollmeier: more idiomatic, but doesn't really matter / make a difference I guess..
```suggestion
    serializedCpg match {
      case Some(serializedCpg) => enhancementExecList.foreach(_.createApplySerializeAndStore(serializedCpg))
      case None =>                enhancementExecList.foreach(_.createAndApply)
    }
```

PR #552, opened 2020-01-20
Added undo functionality to the diffgraph

lifecoder:


codepropertygraph/src/main/scala/io/shiftleft/codepropertygraph/cpgloading/CpgOverlayLoader.scala
```diff
@@ -46,17 +48,35 @@ private[cpgloading] object CpgOverlayLoader {
 private class CpgOverlayApplier(graph: ScalaGraph) {
 
   private val overlayNodeIdToSrcGraphNode: mutable.HashMap[Long, Vertex] = mutable.HashMap()
+  private var inverseBuilder: DiffGraph.InverseBuilder = _
 
   /**
     * Applies diff to existing (loaded) OdbGraph
     */
   def applyDiff(overlay: CpgOverlay): Unit = {
+    inverseBuilder = makeInverseBuilder(false)
```
mpollmeier: to understand the meaning one has to also read `makeInverseBuilder`, because it's unclear what that boolean argument stands for. it's only a few lines below, but i'd still prefer something like
a) inline it, i.e. replace this line  with `inverseBuilder = DiffGraph.InverseBuilder.noop`
b) use named argument

codepropertygraph/src/main/scala/io/shiftleft/codepropertygraph/cpgloading/CpgOverlayLoader.scala
```diff
@@ -46,17 +48,35 @@ private[cpgloading] object CpgOverlayLoader {
 private class CpgOverlayApplier(graph: ScalaGraph) {
 
   private val overlayNodeIdToSrcGraphNode: mutable.HashMap[Long, Vertex] = mutable.HashMap()
+  private var inverseBuilder: DiffGraph.InverseBuilder = _
```
mpollmeier: `var =:= a few months in hell`
just kidding

codepropertygraph/src/test/scala/io/shiftleft/codepropertygraph/cpgloading/DiffGraphTest.scala
```diff
@@ -0,0 +1,105 @@
...
+        SetEdgeProperty(_, EdgeKeyNames.LOCAL_NAME, "old edge attr"), // restore old edge property value
+        SetNodeProperty(_, NodeKeyNames.CODE, "old y code"), // restore old Y property value
+        RemoveNodeProperty(_, NodeKeyNames.ORDER), // remove newly added property
+        RemoveEdge(_), // remove x -> a
+        RemoveEdge(_), // remove b -> c
+        RemoveEdge(_), // remove a -> b
+        RemoveNode(_), // remove c
+        RemoveNode(_), // remove b
+        RemoveNode(_) // remove a
+      ) = changes
```
mpollmeier: does this test assert anything?

codepropertygraph/src/main/scala/io/shiftleft/passes/CpgPass.scala
```diff
@@ -121,6 +121,7 @@ object CpgPass {
   * diff graph, which additionally provides a map from nodes to graph ids.
   * */
 case class AppliedDiffGraph(diffGraph: DiffGraph,
+                            inverseDiffGraph: Option[DiffGraph],
```
fabsx00: We are now storing a `diffGraph` along with the `inverseDiffGraph`. Is this necessary or would it be possible to generate the `inverseDiffGraph` on the fly in the undo operation? I can imagine you store the inverseDiffGraph here so that you can see what was in the properties before the operation, is that why it's done in this way?

codepropertygraph/src/main/scala/io/shiftleft/codepropertygraph/cpgloading/CpgOverlayLoader.scala
GlassAndOneHalf: Could the `InverseBuilder` be passed as an argument to the functions that use it?

codepropertygraph/src/main/scala/io/shiftleft/codepropertygraph/cpgloading/CpgOverlayLoader.scala
```diff
@@ -46,17 +48,35 @@ private[cpgloading] object CpgOverlayLoader {
...
+    addEdgeProperties(overlay)
+    inverseBuilder.build()
+  }
+
+  private def makeInverseBuilder(undoable: Boolean) = {
+    if (undoable)
+      DiffGraph.InverseBuilder.newBuilder
+    else
+      DiffGraph.InverseBuilder.noop
+  }
```
GlassAndOneHalf: I would argue that this could be removed completely and inlined in the corresponding functions

codepropertygraph/src/main/scala/io/shiftleft/codepropertygraph/cpgloading/CpgOverlayLoader.scala
lifecoder: @GlassAndOneHalf yes, and would avoid var, so I'll do that

codepropertygraph/src/main/scala/io/shiftleft/codepropertygraph/cpgloading/CpgOverlayLoader.scala
lifecoder: Yes, this would be more readable, thanks

codepropertygraph/src/main/scala/io/shiftleft/codepropertygraph/cpgloading/CpgOverlayLoader.scala
lifecoder: It could but it would make 8+ duplicated if statements

codepropertygraph/src/main/scala/io/shiftleft/passes/CpgPass.scala
lifecoder: I don't need original diffgraph, this is just for returning multiple values from the function.
It is possible to generate undo from original DiffGraph but then it would need to store old property values and edge references in vars which are set when it is applied which would make it a bit more complicated.

codepropertygraph/src/test/scala/io/shiftleft/codepropertygraph/cpgloading/DiffGraphTest.scala
lifecoder: @mpollmeier Yes, the pattern matching will throw an exception if the structure of `changes` var does not correspond to the pattern above. This could be done with asserts and instanceOf checks of each list element but I think this is more readable

codepropertygraph/src/main/scala/io/shiftleft/passes/CpgPass.scala
fabsx00: So, this means that when we call `applyDiff` with `undoable = true`, we end up with the original diff graph in `diffGraph` and its inverse in `inverseDiffGraph`, correct? That means, we hold the diffgraph along with its inverse in memory at once?

lifecoder: Yes, but you should discard original diffgraph if it is no longer needed and just take the appliedDiff.inverseDiffGraph to keep it in memory until you will need the undo functionality.

fabsx00: Ah, I see, yes, that makes sense. And we'll always have the situation where we briefly hold both, that's something that we won't get around. Ok, approved, from my side.

codepropertygraph/src/test/scala/io/shiftleft/codepropertygraph/cpgloading/DiffGraphTest.scala
mpollmeier: ah right, ok!

codepropertygraph/src/main/scala/io/shiftleft/codepropertygraph/cpgloading/CpgOverlayLoader.scala
mpollmeier: @GlassAndOneHalf see also https://github.com/ShiftLeftSecurity/codepropertygraph/pull/552#discussion_r368772969

PR #553, opened 2020-01-20
Simpler version of #550

fabsx00:
I realized that it makes more sense to check in `createApplySerializeAndStore` whether the serializedCpg is unbacked by a file, in which case we call `createAndApply` instead. This has the same effect as #550 (which is revert by this PR), but with much fewer changes to the code. It also already extends to all passes in `codescience`.


fabsx00:
Tested with `codescience`.

PR #555, opened 2020-01-22
WIP - fix inconsistency in cpg hierarchy: Expression and ArgumentIndex

mpollmeier:


mpollmeier:
There appears to be an inconsistency in our steps hierarchy:
```
trait ExpressionBase[NodeType <: nodes.Expression] 
  extends ArgumentIndexAccessors[NodeType]
```
This doesn't seem to make sense: not every `Expression` node has the `ArgumentIndex` property

The cpg.json doesn't define any `ARGUMENT` nodes, instead it's modelled as an edge, and if we follow those, we end up at the following node types:
```
cpg.argument.raw.label.toSet //"UNKNOWN", "IDENTIFIER", "LITERAL", "CALL" 
```
n.b. `cpg.argument` is equivalent to `cpg.call.argument`

Those nodes however *do all have* an `ARGUMENT_INDEX` property:
```
cpg.argument.raw.count.head //3079
cpg.argument.raw.has(NodeKeys.ARGUMENT_INDEX).count.head //3079
```

In this PR I naively changed the returned Steps type from `Expression` to `NodeSteps[nodes.StoredNode with HasArgumentIndex]` - mostly in lack of a better type. This works in most places, however there's currently two issues, since apparently we also guarantee that those nodes have the `CODE` property, as well as that it's an AST node:
```
[error] codepropertygraph/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/NodeTypeStarters.scala:184:14: value code is not a member of io.shiftleft.semanticcpg.language.NodeSteps[io.shiftleft.codepropertygraph.generated.nodes.StoredNode with io.shiftleft.codepropertygraph.generated.nodes.HasArgumentIndex]
[error]     argument.code(code)
[error]              ^
[error] codepropertygraph/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/operatorextension/package.scala:16:24: value ast is not a member of io.shiftleft.codepropertygraph.generated.nodes.StoredNode with io.shiftleft.codepropertygraph.generated.nodes.HasArgumentIndex
[error]       call.argument(2).ast.isIdentifier
[error]                        ^
```

@ml86 how should we model this best? Should we introduce another nodeBaseTrait? At this point it looks like we need to `extends AST_NODE` and `hasKeys = ["ARGUMENT_INDEX", "CODE"]`, but I may be missing something. Thoughts? What's a good name for this? `ARGUMENT_NODE`?

Likely related to https://github.com/ShiftLeftSecurity/product/issues/3607

ml86:
The problem is that we put the `ARGUMENT_INDEX` in the node used as argument. But we should rather put this information on the `ARGUMENT` edge.
Your suggestion of adding another abstraction `ARGUMENT_NODE` has the same problem as we now have with `EXPRESSION`, which is nodes in this group are not always arguments. For example `BLOCK` is such a node.
So please do not go in the direction you proposed.

mpollmeier:
Ok, so would it be best to change the schema and add the `INDEX` property to the `ARGUMENT` edges? I guess it does involve patching all frontends, but it seems like a very simple fix. The `ARGUMENT_INDEX` property on nodes can just be deprecated in a first step before we later remove it.

ml86:
Yes that would be the fix. But i do not see this to be very urgent.

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/operatorextension/package.scala
```diff
@@ -1,25 +1,28 @@
...
 
   implicit class ArrayAccessExt(val call: Call) extends AnyVal {
 
-    def array: Expression = call.argument(1)
+    def array: StoredNode with HasArgumentIndex =
+      call.argument(1)
 
-    def subscripts: Identifier = call.argument(2).ast.isIdentifier
+    def subscripts: Identifier =
+      call.argument(2).ast.isIdentifier
```
bbrehm: FWIW, this can return either an identifier or a literal. I know you didn't change it, I'm just noticing this old quirk.

PR #559, opened 2020-01-23
use steps via implicit conversion rather than inheritance

mpollmeier:
Note: previously the steps allowed to overload the extended steps, e.g.
`.method` was defined in `ExpressionBase` and was extended in
`Call`. Fortunately the implicit resolution is smart enough to 
unambiguously resolve the correct (same) step, as the following simplified
snippet demonstrates:

```
import scala.language.implicitConversions
trait Base class Child
extends Base

class BaseExt[A <: Base] {
  def foo: String = "BaseExt"
} class ChildExt {
  def foo: String = "ChildExt"
}

implicit def toBaseExt[A <: Base](list: List[A]): BaseExt[A] = new
BaseExt[A]
implicit def toChildExt(list: List[Child]): ChildExt = new
ChildExt

List.empty[Base].foo //BaseExt 
List.empty[Child].foo //ChildExt
```

also 
* deletes unused and incorrect Modifier step
* remove empty Returns steps
* remove empty Declaration step


ml86:
Its good that the compiler tries to pick the most specific implicit conversion but i think most off the time that brings us only a little speed up. Consider this actually compiling snippet:
```
object Pimp extends App {
  class Base
  class Child extends Base

  class BaseExt[A <: Base] {
    def foo: String = "BaseExt"
  }

  class ChildExt[A <:  Child] extends BaseExt {
    override def foo: String = "ChildExt"
  }

  implicit def toBaseExt[A <: Base](list: List[A]): BaseExt[A] = new
      BaseExt[A]
  implicit def toChildExt[A <: Child](list: List[A]): ChildExt[A] = new
      ChildExt


  println(List.empty[Base].foo) //BaseExt
  println(List.empty[Child].foo) //ChildExt
}


In `BaseExt.foo` we need to match on the actual type of the node whereas in `ChildExt.foo` we do not need to do that. In order to do not duplicate the implementation, after matching and figuring out that BaseExt.foo got a Child node, it should call ChildExt.foo.

The second point i want to make is that we should ALWAYS use the form:
class xxxExt[A <: xxx]  {}

Regardless of a type having other types which derive from it we should use this because that way the xxxExt class definitions all have the same style and we can already write the step methods in a way that we do not need to touch the signature if we decide to make derive from a certain type.
In a more practical example `foo` would not return String but, in most cases, a pipe. Rather than fixing the element type of the pipe to `Child` we can this way always use `A`.

**Comment by mpollmeier** [comment]
> Consider this actually compiling snippet

fwiw I just fixed my snippet, it was missing the top two lines, sorry :( That one you can actually paste in a repl.

> Its good that the compiler tries to pick the most specific implicit conversion but i think most off the time that brings us only a little speed up

The point I tried to make isn't about performance, but about correctness. This PR changes the extension method from inheritance to implicits for e.g. Call and Literal, both which used to `override def method` from ExpressionBase. Now they are separate, yet implicit resolution still chooses the correct implementation. 


> In `BaseExt.foo` we need to match on the actual type of the node whereas in `ChildExt.foo` we do not need to do that. In order to do not duplicate the implementation, after matching and figuring out that BaseExt.foo got a Child node, it should call ChildExt.foo.

I don't follow/understand. `BaseExt.foo` doesn't match on anything, it's simply defined as `def foo: String = "BaseExt"`..? 
Also, `BaseExt` may not even know that `ChildExt` exists, because it may be defined in a different compilation unit. 

> we should ALWAYS use the form: `class xxxExt[A <: xxx]  {}`

We currently have two ways of implementing these implicit extensions, and I'd like to only use one of them everywhere, IMO the the simpler one since you only ever need one implicit conversion:
```scala
class Method(raw: GremlinScala[nodes.Method]) extends NodeSteps[nodes.Method]
class MethodExt(raw: GremlinScala[nodes.Method]) extends NodeSteps[nodes.Method]

implicit def toMethod(steps: Steps[nodes.Method]): Method = new Method(steps.raw)
implicit def toMethodExt(steps: Steps[nodes.Method]): MethodExt = new MethodExt(steps.raw)
```
If I understand you correctly, you'd like to continue us using the following model:
```scala
class Method(raw: GremlinScala[nodes.Method]) extends NodeSteps[nodes.Method]
class MethodExt(steps: Method) extends NodeSteps[nodes.Method]

implicit def toMethod(steps: Steps[nodes.Method]): Method = new Method(steps.raw)
implicit def toMethodExt(steps: Method): MethodExt = new MethodExt(steps)
```
I'd like to remove that style and instead use the first style everywhere, because this one requires two implicit conversions to get to `xxxExt`. Does that make sense?

mpollmeier:
@ml86 and I settled this via a call, we're actually on the same boat with all of this. will add a small tweak and merge

PR #566, opened 2020-02-10
refactor ReachingDefPass: increase readability, use specific types where possible

mpollmeier:


dataflowengine/src/main/scala/io/shiftleft/dataflowengine/passes/reachingdef/ReachingDefPass.scala
```diff
@@ -5,31 +5,30 @@ import java.nio.file.Paths
...
 import io.shiftleft.passes.{CpgPass, DiffGraph, ParallelIteratorExecutor}
 import io.shiftleft.semanticcpg.utils.{ExpandTo, MemberAccess}
 import io.shiftleft.semanticcpg.language._
 
+import scala.collection.mutable
 import scala.jdk.CollectionConverters._
 
 class ReachingDefPass(cpg: Cpg) extends CpgPass(cpg) {
-  var dfHelper: DataFlowFrameworkHelper = _
+  lazy val dfHelper: DataFlowFrameworkHelper = new DataFlowFrameworkHelper(cpg.graph)
```
ml86: Please do not make this lazy. 
1. The Helper class is super light weight.
2. There is no reason to believe that a ReachingDefPass is often time created but run() is not called.

mpollmeier: fully agree, not sure why i even did that. change pushed, please re-review.

PR #567, opened 2020-02-11
WIP add edge cardinalities

mpollmeier:
If we statically know the edge cardinality (1:1, 1:n, n:1, n:n) we can
generate more specific code, and e.g. our neighbor accessors could
return a `nodes.Xyz` instead of an `Iterator[nodes.Xyz]`. There are many
areas where this knowledge is already baked into the usage of the cpg,
and moving it into the schema improves readability, performance and
correctness.

This is just a start in order to get initial feedback on the idea. To
obtain the cardinalities I loaded the helloshiftleft cpg/sp and
performed groupCounts on the nodes/edge combinations:
```scala
def cardinalityAnalysis(nodeType: String, edgeType: String) = {
  def nodesTrav = cpg.graph.V.hasLabel(nodeType)
  val nodes = nodesTrav.toList
  val outNodes = nodesTrav.out(edgeType).toList

  println("out cardinalities:")
  println(nodes.groupBy(_.outE(edgeType).count.head).mapValues(_.size).toList.sortBy(_._1))

  println("in cardinalities:")
  outNodes.groupBy(_.label).foreach { case (label, node) =>
    println(label + ": " + node.groupBy(_.inE(edgeType).count.head).mapValues(_.size).toList.sortBy(_._1))
  }
}

cardinalityAnalysis("FILE", "CFG")
cardinalityAnalysis("METHOD", "AST")
cardinalityAnalysis("METHOD", "CFG")
cardinalityAnalysis("TYPE", "AST")
cardinalityAnalysis("TYPE_DECL", "AST")
cardinalityAnalysis("TYPE_DECL", "BINDS")
cardinalityAnalysis("TYPE_DECL", "VTABLE")
```

### DIFF
--- a/codepropertygraph/src/main/resources/schemas/base.json
+++ b/codepropertygraph/src/main/resources/schemas/base.json
@@ -72,7 +72,7 @@
          "keys": ["NAME", "ORDER"],
          "comment": "Node representing a source file. Often also the AST root",
          "outEdges": [
-             {"edgeName": "AST", "inNodes": ["NAMESPACE_BLOCK"]}
+           {"edgeName": "AST", "inNodes": ["NAMESPACE_BLOCK"], "cardinality": "oneToOne"}
          ],
 	 "is" : ["AST_NODE"]
         },
@@ -86,8 +86,11 @@
          "is": ["DECLARATION", "CFG_NODE", "AST_NODE"],
          "outEdges" : [
              {"edgeName": "AST", "inNodes": ["METHOD_RETURN", "METHOD_PARAMETER_IN",
-                                             "MODIFIER", "BLOCK", "TYPE_PARAMETER"]},
-             {"edgeName": "CFG", "inNodes": ["CALL", "IDENTIFIER", "FIELD_IDENTIFIER", "LITERAL", "METHOD_REF", "METHOD_RETURN", "RETURN", "BLOCK", "UNKNOWN"]}
+                                             "MODIFIER", "BLOCK", "TYPE_PARAMETER"],
+              "cardinality": "oneToMany"},
+             {"edgeName": "CFG", 
+              "inNodes": ["CALL", "IDENTIFIER", "FIELD_IDENTIFIER", "LITERAL", "METHOD_REF", "METHOD_RETURN", "RETURN", "BLOCK", "UNKNOWN"],
+              "cardinality": "oneToOne" }
          ]
         },
 
@@ -122,8 +125,8 @@
          "keys" : ["NAME", "FULL_NAME", "IS_EXTERNAL", "INHERITS_FROM_TYPE_FULL_NAME", "AST_PARENT_TYPE", "AST_PARENT_FULL_NAME", "ALIAS_TYPE_FULL_NAME", "ORDER"],
          "comment" : "A type declaration",
          "outEdges" : [
-             {"edgeName": "AST", "inNodes": ["TYPE_PARAMETER", "MEMBER", "MODIFIER"]},
-             {"edgeName": "BINDS", "inNodes": ["BINDING"]},
+             {"edgeName": "AST", "inNodes": ["TYPE_PARAMETER", "MEMBER", "MODIFIER"], "cardinality": "oneToMany"},
+             {"edgeName": "BINDS", "inNodes": ["BINDING"], "cardinality": "oneToMany"},
              {"edgeName": "VTABLE", "inNodes": ["METHOD"]}
          ],
 	 "is" : ["AST_NODE"]


ml86:
I like the general idea.
Problem is that in this PR in the json files you only encode the outgoing edge cardinalities.
If we want to introduce the cardinalities we should also encode them for the incoming edges like we do here: https://github.com/ShiftLeftSecurity/codepropertygraph/blob/master/cpgvalidator/src/main/scala/io/shiftleft/cpgvalidator/facts/InFactsImporter.scala and here: https://github.com/ShiftLeftSecurity/codepropertygraph/blob/master/cpgvalidator/src/main/scala/io/shiftleft/cpgvalidator/facts/OutFactsImporter.scala

mpollmeier:
Ah, good to know that we already have those documented, that'll help validate these changes. Once this is complete, those are probably not needed any more, because the schema doesn't allow to get those cardinalities wrong any longer. 

The edge cardinalities are only specified at the outgoing edges, but they define the relationship both ways. Examples for a given `A --- someEdge ---> B`:
* `A.outEdge cardinality=1:1` -> `A.someEdgeOut: B` && `B.someEdgeIn: A`
* `A.outEdge cardinality=1:n` -> `A.someEdgeOut: Iterator[B]` && `B.someEdgeIn: A`
* `A.outEdge cardinality=n:1` -> `A.someEdgeOut: B` && `B.someEdgeIn: Iterator[A]`
* `A.outEdge cardinality=n:n` -> `A.someEdgeOut: Iterator[B]` && `B.someEdgeIn: Iterator[A]`

Does this make sense, and/or am I missing something?

ml86:
It is not quite so easy. Consider this: `A --- someEdge --> C` and `B --- someEdge --> C`. Both have 1:1 relations. In that case the signature of `C.someEdgeIn` still needs to be `Iterator[Ancestor of A and B]`.

I am not sure whether this is something we want to pursue at the moment.

mpollmeier:
Reawakening this old beauty... 
Since the initial proposal you convinced me that the new (non-underscore) neighbor accessors (that I added) should not contain the edge names, since those are implementation details. 

Note that your last comment assumes a different interpretation of what that `someEdgeIn` neighbours should do. It's a good thought, but for this I'd like to only look at direct neighbours.

Also, we agreed in https://github.com/ShiftLeftSecurity/codescience/pull/3469 that differentiating three cardinalities would be benefitial: `one`, `zeroOrOne`, `many`. Those would be reflected in the types `T`, `Option[T]` and `List[T]`. Since we define only `out` edges in our schema (the `in` edges are implicit), we need to define the cardinality for both sides on the `out` edge.

Based on all that, here's an updated version of the cases I'd like to handle, given two nodes `A` and `B`:
| A.outEdge(inNode=B) cardinality | A.b type    | B.a type    |
| ------------------------------- | ------------| ------------|
| `1:1`                           | `B`         | `A`         | 
| `1:0-1`                         | `Option[B]` | `A`         |
| `0-1:1`                         | `B`         | `Option[A]` |
| `1:n`                           | `List[B]`   | `A`         |
| `n:1`                           | `B`         | `List[A]`   |
| `0-1:n`                         | `List[B]`   | `Option[A]` |
| `n:0-1`                         | `Option[B]` | `List[A]`   |


Notes:
1) the edge name is now transparent to the user, which is intended because it's an implementation detail that may change
2) the local names `A.b` and `B.a` will be customisable in the schema
3) some of those are probably never used

ml86:
I like encoding the relations in the JSON files.
I am not sure whether you understood my last comment as it was intended.
I wanted to express:
If you have A -> C with a 1:1 relation and B -> C with a 1:1 relation and in both cases we are talking about AST edges, than the AST-IN relation starting at C need to return List[Ancestor of A and B].
Or is you intention of creating accessors like AST-IN-TO-A and AST-IN-TO-B?

mpollmeier:
Ah ok I think I understand what you mean now. We're talking about two concepts that are closely related, but semantically different. What I didn't explicitly state is that we will still have the `_astIn` (with or without underscore prefix - that's a separate discussion) neighbour accessors, and we'll actually use them here. Example schema:
```
A --- AST --> C
B --- AST --> C
cardinality 1:1 for both
```

Generated node source:
```
class C {
  def _astIn: Iterator[StoredNode] = createAdjacentNodeIteratorByOffSet(...)
  def a: A = _astIn.collect { case a: A => a }.head
  def b: B = _astIn.collect { case b: B => b }.head
}

// for A/B we can start off using the same simple model
class A {
  def _astOut: Iterator[StoredNode] = createAdjacentNodeIteratorByOffSet(0)
  def c: C = _astOut.head
}

// and - maybe in a follow up refactoring - use an optimization for the 'non-list' cases, like this:
class A {
  def _astOut: Iterator[StoredNode] = createAdjacentNodeIteratorByOffSet(0)
  
  /* does not use `_astOut` or instantiate any other iterator,
   * instead uses direct pointer onto the array entry */
  def c: C = getTheOneAndOnlyByOffSet(0)
}
```

ml86:
I dont think it is a good idea to put the methods for `a`, `b`, `c` without a leading  _. It will severely hinder us in defining the DSL. Just because we have an edge from A -> B we do not always want a method named `b` on all a vertices. So please also put an _ infront of those methods.

mpollmeier:
Ok. I was assuming that we want to make all these connections easily accessible and manage their construction automatically. 

But maybe that's not the case and it's better to maintain them manually. I'll still generate them as above, just prefix them with `_` to indicate that they're not part of the official DSL. And the manually maintained node extensions can then just wrap them, e.g.
```scala
class CExt(wrapped: nodes.C) {
  def a: A = wrapped._a
```

@fabsx00 do you agree with @ml86 ?

mpollmeier:
superseded shortly

PR #572, opened 2020-02-12
comment.file: handle in NodeSteps for safety and clarity

mpollmeier:
Before this PR, whether we use the correct traversal depended on the
usage when calling `.file`. If we statically only know that we have a
`NodeSteps[nodes.StoredNodes]`, they will *not* use the more specific
`Comment.file`, even if the nodes all actually *are* of type
`nodes.Comment`.

Therefor it's safer (and easier to reason about) if we make the
distinction on NodeSteps. The obvious downside is that we clutter
`Comment` related things in `NodeSteps`, but that was already the case
for Namespace.


fabsx00:
Approved for now as for `comment`, it's alright. Let's make sure though that we do not sacrifice decomposition of the language by introducing higher-level node types in `NodeSteps` of `semanticcpg`.

mpollmeier:
Yes. Our DSL has many advantages, this one is an obvious disadvantage. We can only add more steps, overriding is not safe. 

One clean :tm: way to do this would be to remove `NodeSteps.file` and instead define it in a mixin `trait WithFile` that has one default implementation and can be overridden. Downside: we need to define a Steps class and it's implicit for every node that does support `.file`. Maybe not too bad? 

I'll merge this for now, let me know your thoughts on the above.

PR #578, opened 2020-02-14
return NodeSteps everywhere, do not extend NodeSteps

mpollmeier:
use AnyVal instead

as discussed in https://github.com/ShiftLeftSecurity/codescience/pull/3377#issuecomment-585432344
part of https://github.com/ShiftLeftSecurity/product/issues/3526

this isn't yet complete, but it doesn't harm if we do it in multiple steps

### DIFF
--- a/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/operatorextension/AssignmentTrav.scala
+++ b/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/operatorextension/AssignmentTrav.scala
@@ -1,16 +1,13 @@
 package io.shiftleft.semanticcpg.language.operatorextension
 
 import gremlin.scala.GremlinScala
-
-import io.shiftleft.codepropertygraph.generated.nodes.{Call, Expression}
+import io.shiftleft.codepropertygraph.generated.nodes
 import io.shiftleft.semanticcpg.language.Steps
 
 /**
   * A wrapper for assignment calls that offers syntactic sugar
   * */
-class AssignmentTrav(override val raw: GremlinScala[Call]) extends Steps[Call](raw) {
-
+class AssignmentTrav(override val raw: GremlinScala[nodes.Call]) extends Steps[nodes.Call](raw) {
   def target: TargetTrav = new TargetTrav(map(_.target).raw)
-
-  def source: Steps[Expression] = map(_.source)
+  def source: Steps[nodes.Expression] = map(_.source)
 }
--- a/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/types/structure/MethodParameter.scala
+++ b/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/types/structure/MethodParameter.scala
@@ -10,15 +10,15 @@ import io.shiftleft.semanticcpg.language.types.propertyaccessors._
 /**
   * Formal method input parameter
   * */
-class MethodParameter(raw: GremlinScala[nodes.MethodParameterIn])
-    extends NodeSteps[nodes.MethodParameterIn](raw)
-    with EvalTypeAccessors[nodes.MethodParameterIn] {
+class MethodParameter(val wrapped: NodeSteps[nodes.MethodParameterIn])
+    extends EvalTypeAccessors[nodes.MethodParameterIn] {
+  override def raw: GremlinScala[nodes.MethodParameterIn] = wrapped.raw
 
   /**
     * Traverse to all `num`th parameters
     * */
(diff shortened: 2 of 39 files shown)


semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/types/structure/MethodParameter.scala
```diff
@@ -10,39 +10,39 @@ import io.shiftleft.semanticcpg.language.types.propertyaccessors._
 /**
   * Formal method input parameter
   * */
-class MethodParameter(raw: GremlinScala[nodes.MethodParameterIn])
-    extends NodeSteps[nodes.MethodParameterIn](raw)
-    with EvalTypeAccessors[nodes.MethodParameterIn] {
+class MethodParameter(val wrapped: NodeSteps[nodes.MethodParameterIn])
+    extends EvalTypeAccessors[nodes.MethodParameterIn] {
```
ml86: Why do we need to extend from EvalTypeAccessors here? There should be a separate pimp class with its on implicit conversion.

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/types/structure/MethodParameterOut.scala
```diff
@@ -7,29 +7,29 @@ import io.shiftleft.semanticcpg.language._
 import io.shiftleft.semanticcpg.language.types.expressions.generalizations.Expression
 import io.shiftleft.semanticcpg.language.types.propertyaccessors._
 
-class MethodParameterOut(raw: GremlinScala[nodes.MethodParameterOut])
-    extends NodeSteps[nodes.MethodParameterOut](raw)
-    with EvalTypeAccessors[nodes.MethodParameterOut] {
+class MethodParameterOut(val wrapped: NodeSteps[nodes.MethodParameterOut])
+    extends EvalTypeAccessors[nodes.MethodParameterOut] {
```
ml86: Same as form MethodParameter.

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/types/structure/TypeDecl.scala
```diff
@@ -10,22 +10,21 @@ import org.apache.tinkerpop.gremlin.structure.Direction
 /**
   * Type declaration - possibly a template that requires instantiation
   * */
-class TypeDecl(raw: GremlinScala[nodes.TypeDecl])
-    extends NodeSteps[nodes.TypeDecl](raw)
-    with ModifierAccessors[nodes.TypeDecl] {
+class TypeDecl(val wrapped: NodeSteps[nodes.TypeDecl]) extends ModifierAccessors[nodes.TypeDecl] {
```
ml86: Same as above from MethodParameter. Why extend from ModifiierAccessors?

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/operatorextension/AssignmentTrav.scala
```diff
@@ -1,16 +1,13 @@
...
-import io.shiftleft.codepropertygraph.generated.nodes.{Call, Expression}
+import io.shiftleft.codepropertygraph.generated.nodes
 import io.shiftleft.semanticcpg.language.Steps
 
 /**
   * A wrapper for assignment calls that offers syntactic sugar
   * */
-class AssignmentTrav(override val raw: GremlinScala[Call]) extends Steps[Call](raw) {
-
+class AssignmentTrav(override val raw: GremlinScala[nodes.Call]) extends Steps[nodes.Call](raw) {
```
ml86: Since this is still the old pattern, i think we should extend from `NodeSteps` instead of `Steps` otherwise we wont have steps like `file`.

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/types/structure/MethodParameter.scala
mpollmeier: agreed, and it's on my list already, but that'll make it into a follow-up PR.

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/operatorextension/AssignmentTrav.scala
mpollmeier: Yup agreed, thanks for spotting. I did that in some other places, however apparently not all. At the end I'll look for all `extends NodeSteps` and `extends Steps` and double check which ones are required.

PR #581, opened 2020-02-17
return NodeSteps everywhere

mpollmeier:


fabsx00:
Merging with @mpollmeier's permission.

PR #584, opened 2020-02-17
clean up implicits for operatorextensions

mpollmeier:


semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/operatorextension/package.scala
```diff
@@ -1,48 +1,74 @@
...
 
 package object operatorextension {
 
-  implicit class ArrayAccessExt(val call: nodes.Call) extends AnyVal {
+  trait Implicits {
+    implicit def toNodeTypeStartersOperatorExtension(cpg: Cpg): NodeTypeStarters = new NodeTypeStarters(cpg)
+    implicit def toArrayAccessExt(call: nodes.Call): ArrayAccessExt = new ArrayAccessExt(call)
+    implicit def toArrayAccessTrav(steps: Steps[nodes.Call]): ArrayAccessTrav = new ArrayAccessTrav(steps)
+    implicit def toAssignmentExt(call: nodes.Call): AssignmentExt = new AssignmentExt(call)
+    implicit def toAssignmentTrav(steps: Steps[nodes.Call]): AssignmentTrav = new AssignmentTrav(steps)
```
ml86: This looks to me like we now have the array and assignment steps on all CALL nodes. Is that really the intention?

fabsx00: No, that would not be intended. @mpollmeier is this the case?

fabsx00: The idea is to have an `isAssignment`, which filters calls for assignments and presents results wrapped in an assignment class, and that that provides methods that are only valid for assignments.

ml86: I suggest we create a `nodes.AssignmentCall` which derives from `nodes.Call` and than provide an implicit conversion to `AssignmentExt` for `NodeSteps[nodes.AssignmentCall]`.

mpollmeier: ok that makes sense, thanks for catching!

PR #585, opened 2020-02-17
Fix performance issue with ReachingDefPass

lifecoder:


fabsx00:
cc @bbrehm @ml86 didn't one of you recently touch this?

PR #587, opened 2020-02-18
streamline extension methods via implicits

mpollmeier:
remaining 'property' accessors

### DIFF
--- a/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/types/expressions/Call.scala
+++ b/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/types/expressions/Call.scala
@@ -13,7 +13,7 @@ import io.shiftleft.semanticcpg.language.types.structure.{Member, Method, Method
   A call site
   */
 class Call(val wrapped: NodeSteps[nodes.Call]) extends AnyVal {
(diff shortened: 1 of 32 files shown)


semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/types/expressions/Call.scala
```diff
@@ -12,8 +12,8 @@ import io.shiftleft.semanticcpg.language.types.structure.{Member, Method, Method
 /**
   A call site
   */
-class Call(val wrapped: NodeSteps[nodes.Call]) extends EvalTypeAccessors[nodes.Call] {
-  override val raw: GremlinScala[nodes.Call] = wrapped.raw
+class Call(val wrapped: NodeSteps[nodes.Call]) extends AnyVal {
+  def raw: GremlinScala[nodes.Call] = wrapped.raw
```
ml86: Make this private. Otherwise we only have a `raw` step for some node types depending on the fact whether we use such a short hand or not.
I have seen this `raw` method not being private in other places too. Please correct those also.

mpollmeier: agreed, done in 05a8965f1af48eadf4b9b5264c8d842d66c2157f

PR #590, opened 2020-02-20
Add FieldIdentifier related steps.

ml86:
- Added step to filter AST nodes for FieldIdentifier.
- Added property accessor for canonicalName property of FieldIdentifier.


AlexDenisov:
Awesome @ml86!

PR #592, opened 2020-02-20
constraints liberate: specific types for assignment/arrayaccess/...

mpollmeier:
add refinements for calls

as discussed in https://github.com/ShiftLeftSecurity/codepropertygraph/pull/584#pullrequestreview-360221290

### DIFF
--- a/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/operatorextension/package.scala
+++ b/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/operatorextension/package.scala
@@ -4,52 +4,57 @@ import gremlin.scala.GremlinScala
 import io.shiftleft.codepropertygraph.Cpg
 import io.shiftleft.codepropertygraph.generated.Operators
 import io.shiftleft.codepropertygraph.generated.nodes
+import io.shiftleft.codepropertygraph.generated.nodes.Call
+import io.shiftleft.overflowdb.OdbGraph
 
 package object operatorextension {
+  class Assignment(call: Call) extends Call(call.graph.asInstanceOf[OdbGraph], call.id)
+  class Arithmetic(call: Call) extends nodes.Call(call.graph.asInstanceOf[OdbGraph], call.id)
+  class ArrayAccess(call: Call) extends nodes.Call(call.graph.asInstanceOf[OdbGraph], call.id)
 
   trait Implicits {
     implicit def toNodeTypeStartersOperatorExtension(cpg: Cpg): NodeTypeStarters = new NodeTypeStarters(cpg)
-    implicit def toArrayAccessExt(call: nodes.Call): ArrayAccessExt = new ArrayAccessExt(call)
-    implicit def toArrayAccessTrav(steps: Steps[nodes.Call]): ArrayAccessTrav = new ArrayAccessTrav(steps)
-    implicit def toAssignmentExt(call: nodes.Call): AssignmentExt = new AssignmentExt(call)
-    implicit def toAssignmentTrav(steps: Steps[nodes.Call]): AssignmentTrav = new AssignmentTrav(steps)
+    implicit def toArrayAccessExt(arrayAccess: ArrayAccess): ArrayAccessExt = new ArrayAccessExt(arrayAccess)
+    implicit def toArrayAccessTrav(steps: Steps[ArrayAccess]): ArrayAccessTrav = new ArrayAccessTrav(steps)
+    implicit def toAssignmentExt(assignment: Assignment): AssignmentExt = new AssignmentExt(assignment)
+    implicit def toAssignmentTrav(steps: Steps[Assignment]): AssignmentTrav = new AssignmentTrav(steps)
     implicit def toTargetExt(call: nodes.Expression): TargetExt = new TargetExt(call)
     implicit def toTargetTrav(steps: Steps[nodes.Expression]): TargetTrav = new TargetTrav(steps)
     implicit def toOpAstNodeExt[A <: nodes.AstNode](node: A): OpAstNodeExt[A] = new OpAstNodeExt(node)
     implicit def toOpAstNodeTrav[A <: nodes.AstNode](steps: Steps[A]): OpAstNodeTrav[A] = new OpAstNodeTrav(steps)
   }
 
-  class ArrayAccessExt(val call: nodes.Call) extends AnyVal {
-    def array: nodes.Expression = call.argument(1)
-    def subscripts: NodeSteps[nodes.Identifier] = call.argument(2).ast.isIdentifier
+  class ArrayAccessExt(val arrayAccess: ArrayAccess) extends AnyVal {
+    def array: nodes.Expression = arrayAccess.argument(1)
+    def subscripts: NodeSteps[nodes.Identifier] = arrayAccess.argument(2).ast.isIdentifier
   }
 
-  class ArrayAccessTrav(val wrapped: Steps[nodes.Call]) extends AnyVal {
-    private def raw: GremlinScala[nodes.Call] = wrapped.raw
+  class ArrayAccessTrav(val wrapped: Steps[ArrayAccess]) extends AnyVal {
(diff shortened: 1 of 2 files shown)


semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/operatorextension/package.scala
```diff
@@ -4,76 +4,82 @@ import gremlin.scala.GremlinScala
 import io.shiftleft.codepropertygraph.Cpg
 import io.shiftleft.codepropertygraph.generated.Operators
 import io.shiftleft.codepropertygraph.generated.nodes
+import io.shiftleft.codepropertygraph.generated.nodes.Call
+import io.shiftleft.overflowdb.OdbGraph
 
 package object operatorextension {
+  class Assignment(call: Call) extends Call(call.graph.asInstanceOf[OdbGraph], call.id)
```
ml86: This should be `extends nodes.Call` instead of `extends Call`.

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/operatorextension/package.scala
```diff
@@ -4,76 +4,82 @@ import gremlin.scala.GremlinScala
...
+  class Arithmetic(call: Call) extends nodes.Call(call.graph.asInstanceOf[OdbGraph], call.id)
+  class ArrayAccess(call: Call) extends nodes.Call(call.graph.asInstanceOf[OdbGraph], call.id)
```
ml86: As long as the Traversal classes do not have names indicating them being traversals, I suggest we but these node types into a `nodes` namespace like the generated nodes and only use them via `nodes.XXX`. That way it is directly visible whether we talk about nodes or traversals.

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/operatorextension/package.scala
```diff
@@ -4,76 +4,82 @@ import gremlin.scala.GremlinScala
...
   }
 
   class OpAstNodeTrav[A <: nodes.AstNode](val wrapped: Steps[A]) extends AnyVal {
     private def raw: GremlinScala[A] = wrapped.raw
-    def inAssignment: NodeSteps[nodes.Call] = wrapped.flatMap(_.inAssignment)
-    def assignments: NodeSteps[nodes.Call] = wrapped.flatMap(_.assignments)
-    def arithmetics: NodeSteps[nodes.Call] = wrapped.flatMap(_.arithmetics)
+    def inAssignment: NodeSteps[Assignment] = wrapped.flatMap(_.inAssignment)
+    def assignments: NodeSteps[Assignment] = wrapped.flatMap(_.assignments)
+    def arithmetics: NodeSteps[Arithmetic] = wrapped.flatMap(_.arithmetics)
```
ml86: The hole OpAstNodeTrav class seems to be a unnecessary. Why are the three methods not defined in AstNode.scala?

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/operatorextension/package.scala
```diff
@@ -4,76 +4,82 @@ import gremlin.scala.GremlinScala
...
                    Operators.indirectComputedMemberAccess,
                    Operators.indexAccess,
                    Operators.indirectIndexAccess)
+        .map(new ArrayAccess(_))
   }
 
   class TargetTrav(val wrapped: NodeSteps[nodes.Expression]) extends AnyVal {
     private def raw: GremlinScala[nodes.Expression] = wrapped.raw
-    def isArrayAccess: NodeSteps[nodes.Call] = wrapped.flatMap(_.isArrayAccess)
+    def isArrayAccess: NodeSteps[ArrayAccess] = wrapped.flatMap(_.isArrayAccess)
```
ml86: Same as for OpAstNodeTrav. The hole class is unnecessary and the methods should go to Expression.scala

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/operatorextension/package.scala
```diff
@@ -4,76 +4,82 @@ import gremlin.scala.GremlinScala
...
     def expr: NodeSteps[nodes.Expression] = wrapped.map(_.expr)
```
ml86: While you are moving the code remove this method. It does nothing.

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/operatorextension/package.scala
fabsx00: We have this class so that all steps related to AST nodes, which only make sense when the operator extension is used are in `operatorextension`. That way, it becomes possible to remove an extension without modifying `ASTNode.scala`. In addition to providing functionality, the operatorextension is used as an example of how to provide an extension. Please leave it like this.

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/operatorextension/package.scala
fabsx00: No, as explained above. This is about augmenting `Expression`.

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/operatorextension/package.scala
fabsx00: This method checks whether the target is an array access.

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/operatorextension/package.scala
ml86: Ok makes sense to keep that separate. I would than suggest to not put the classes in question into the package.scala in order to keep the same layout as for the rest of the AST steps.

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/operatorextension/package.scala
ml86: Same as above.

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/operatorextension/package.scala
ml86: Sorry this was not clear. I meant the `expr` method.

fabsx00: aah, ok, yes.

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/operatorextension/package.scala
mpollmeier: I agree, but for whatever reason scalac needs it this way. I guess it's a weird edge case when using relative imports inside a package object, or whatever. I could debug this but don't think it's worth it, esp. given that we may rethink the naming of steps/nodes shortly.

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/operatorextension/package.scala
mpollmeier: Good idea, will do

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/operatorextension/package.scala
mpollmeier: fwiw here's the scalac error, edge case doesn't seem to be the package object though. 
```
Error:(12, 26) stable identifier required, but Assignment.this.nodes found.
  class Assignment(call: nodes.Call) extends nodes.Call(call.graph.asInstanceOf[OdbGraph], call.id)
```

PR #599, opened 2020-02-25
move `.tag` into implicit extension, to allow more specific extensions

mpollmeier:


semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/package.scala
```diff
@@ -243,6 +243,16 @@ package object language extends operatorextension.Implicits {
       new Steps[NodeType](__[NodeType](iter.to(Seq): _*))
   }
 
+  implicit class NodeStepsExt(steps: Steps[_ <: StoredNode]) {
```
ml86: Why is this not an implicit def + a class which extends AnyVal?

mpollmeier: I just added `extends AnyVal` (and did the same for all implicit classes in the same file). 

But I kept `implicit class` rather than `implicit def`, because it's less boilerplate. The idea is that we only use `implicit def` if the extension class is defined in another file. That way we can keep all the implicit conversions in one place, but still define larger classes in separate files. 

If you prefer to have more boilerplate for gaining more consistency, I don't mind changing it - let me know.

mpollmeier:
merged as part of https://github.com/ShiftLeftSecurity/codepropertygraph/pull/604
@ml86 let me know if you want to respond to https://github.com/ShiftLeftSecurity/codepropertygraph/pull/599#discussion_r384225191

PR #608, opened 2020-02-28
use typeclasses for pretty printing

mpollmeier:
We override `.p` in a few different steps, always relying on
overriding it in a more specific `Steps` class. This is rather clumsy
and doesn't fit well with our new streamlined implicit extension
design. This PR proposes to replace it with type classes. The tests
show all use cases we have for `.p`.

This is mostly a RFC PR. If we like this pattern, I'll
replace the old `.p` everywhere.

Related: https://github.com/ShiftLeftSecurity/codescience/issues/3406

Part of https://github.com/ShiftLeftSecurity/product/issues/3526


mpollmeier:
> I think the print functions should be defined on the individual nodes like the other expansions and the traversals just refer to them

Note that this propsal already does that, just not in the traditional object oriented way: `Show` is a typeclass instance for a given Node (and not for a Traversal). So the `Show` typeclasses are already defined per node, and the traversals 'refer' to them, as the tests demonstrate. 

AFAICS the simple model you propose doesn't really work for us, because `nodes.Xyz` and `Steps.p` are defined here, while the implementations of `toString` are defined in different compilation units, e.g. https://github.com/ShiftLeftSecurity/codescience/blob/cad60c8/securityprofile/src/main/scala/io/shiftleft/securityprofile/language/steps/FindingStep.scala#L41 and https://github.com/ShiftLeftSecurity/codescience/blob/cad60c88ed8064ac8ba00fa30a5e08facd590bba/dataflowengine/src/main/scala/io/shiftleft/dataflowengine/language/Flow.scala#L18

I don't currently see a better model that doesn't rely on explicitly returning and passing around `FindingStep` instances, which goes against our new streamlined model of just returning and passing `NodeSteps` everywhere, and then using implicit extensions via value classes.

ml86:
Lets have a zoom about this. Ping me on slack please.

fabsx00:
You reference this PR in https://github.com/ShiftLeftSecurity/codescience/issues/3406, but I'm not sure it's related, as that one talks about the way `containedNodes` are displayed on the shall when invoking `.l`, and that seems an issue different o how specializations of `.p` are created. Am I missing something?

mpollmeier:
@ml86 as discussed via zoom I went ahead with this model. The changeset is small and includes the streamlining of `Flows`. Please re-review. 
After your feedback I'll provide a CS PR before merging this one.

mpollmeier:
btw the changes in cpgserver/CpgRouteSpec.scala are just scalafmt. you can ignore that one, I didn't make any actual changes

PR #611, opened 2020-03-01
refactor: simplify .method extension step

mpollmeier:
random refactorings as I'm going through the codebase to get a better
understanding


dataflowengine/src/main/scala/io/shiftleft/dataflowengine/language/nodemethods/TrackingPointMethods.scala
```diff
@@ -16,13 +15,12 @@ class TrackingPointMethods(val node: nodes.TrackingPointBase) extends AnyVal {
   /**
     * Convert to nearest AST node
     * */
-  def astNode: nodes.AstNode = {
+  def astNode: nodes.AstNode =
```
ml86: Please stop removing "unnecessary" brackets. This just creates noise and is personal preference.

mpollmeier: it's just itching me when i see those... ok then :)

PR #612, opened 2020-03-01
use new typesafe neighbor accessors

mpollmeier:
* part of migration to new DSL:
https://github.com/ShiftLeftSecurity/product/issues/3526
* faster than tinkerpop api
* more concise and readable
* (far) less casts

### DIFF
--- a/dataflowengine/src/main/scala/io/shiftleft/dataflowengine/language/TrackingPoint.scala
+++ b/dataflowengine/src/main/scala/io/shiftleft/dataflowengine/language/TrackingPoint.scala
@@ -5,7 +5,6 @@ import io.shiftleft.codepropertygraph.generated._
 import io.shiftleft.semanticcpg.language.{NodeSteps, Steps}
 import io.shiftleft.Implicits.JavaIteratorDeco
 import io.shiftleft.semanticcpg.utils.{ExpandTo, MemberAccess}
-import org.apache.tinkerpop.gremlin.structure.Vertex
 import scala.jdk.CollectionConverters._
 
 /**
@@ -99,9 +98,7 @@ class TrackingPoint(val wrapped: NodeSteps[nodes.TrackingPoint]) extends AnyVal 
 
   private def indirectAccess(node: nodes.StoredNode): Boolean =
     node match {
-      case call: nodes.Call =>
-        val callName = call.value2(NodeKeys.NAME)
-        MemberAccess.isGenericMemberAccessName(callName)
+      case call: nodes.Call => MemberAccess.isGenericMemberAccessName(call.name)
       case _ => false
     }
 
--- a/dataflowengine/src/main/scala/io/shiftleft/dataflowengine/language/TrackingPoint.scala
+++ b/dataflowengine/src/main/scala/io/shiftleft/dataflowengine/language/TrackingPoint.scala
@@ -99,7 +99,7 @@ class TrackingPoint(val wrapped: NodeSteps[nodes.TrackingPoint]) extends AnyVal 
   private def indirectAccess(node: nodes.StoredNode): Boolean =
     node match {
       case call: nodes.Call => MemberAccess.isGenericMemberAccessName(call.name)
-      case _ => false
+      case _                => false
     }
 
 }
--- a/dataflowengine/src/main/scala/io/shiftleft/dataflowengine/language/TrackingPoint.scala
+++ b/dataflowengine/src/main/scala/io/shiftleft/dataflowengine/language/TrackingPoint.scala
@@ -94,7 +94,7 @@ class TrackingPoint(val wrapped: NodeSteps[nodes.TrackingPoint]) extends AnyVal 
         case NodeTypes.METHOD_PARAMETER_OUT =>
           ExpandTo.parameterInToMethod(dataFlowObject)
         case NodeTypes.LITERAL | NodeTypes.CALL | NodeTypes.IDENTIFIER | NodeTypes.RETURN | NodeTypes.UNKNOWN =>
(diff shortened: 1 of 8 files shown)


semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/types/structure/TypeDecl.scala
```diff
@@ -112,21 +111,12 @@ class TypeDecl(val wrapped: NodeSteps[nodes.TypeDecl]) extends AnyVal {
...
-          .vertices(Direction.OUT, EdgeTypes.REF)
-          .next
-          .asInstanceOf[nodes.TypeDecl]
-      } else {
-        typeDecl
-      }
-    })
-  }
+        typeDecl.aliasOfOut.next.refOut.next
+      } else typeDecl
```
ml86: If the `if` block has bracket PLEASE do also use brackets for the `else` block. This is just asking for bugs.

dataflowengine/src/main/scala/io/shiftleft/dataflowengine/language/TrackingPoint.scala
```diff
@@ -84,26 +83,22 @@ class TrackingPoint(val wrapped: NodeSteps[nodes.TrackingPoint]) extends AnyVal
...
-          ExpandTo.parameterInToMethod(dataFlowObject)
-        case NodeTypes.METHOD_PARAMETER_OUT =>
-          ExpandTo.parameterInToMethod(dataFlowObject)
-        case NodeTypes.LITERAL | NodeTypes.CALL | NodeTypes.IDENTIFIER | NodeTypes.RETURN | NodeTypes.UNKNOWN =>
-          ExpandTo.expressionToMethod(dataFlowObject)
-      }
-    method.asInstanceOf[nodes.Method]
-  }
+  // TODO match on type rather than label: depends on https://github.com/ShiftLeftSecurity/codepropertygraph/issues/610
+  private def methodFast(dataFlowObject: nodes.TrackingPoint): nodes.Method =
```
bbrehm: We could 
```
match{
      dataFlowObject case methodReturn: nodes.MethodReturn =>   ExpandTo.methodReturnToMethod(methodReturn)
      case methodParameterIn: nodes.MethodParameterIn =>  ExpandTo.parameterInToMethod(methodParameterIn)
      case methodParameterOut: nodes.MethodParameterOut => ExpandTo.parameterOutToMethod(methodParameterOut)
      case expression: nodes.Expression =>  ExpandTo.expressionToMethod(expression)
      case _ => throw new NotImplementedError() //e.g. free tracking point
    }
```

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/utils/ExpandTo.scala
```diff
@@ -17,91 +17,74 @@ object ExpandTo {
...
-  def callReceiver(callNode: Vertex): nodes.StoredNode = callReceiverOption(callNode).get
+  def callReceiver(callNode: nodes.Call): nodes.StoredNode =
+    callReceiverOption(callNode).get
 
   // TODO move into traversal dsl - used in codescience
-  def callArguments(callNode: Vertex): Iterator[nodes.Expression] = {
-    callNode.asInstanceOf[nodes.StoredNode]._argumentOut.asScala.map(_.asInstanceOf[nodes.Expression])
-  }
+  def callArguments(callNode: nodes.CallRepr): Iterator[nodes.Expression] =
+    callNode._argumentOut.asScala.map(_.asInstanceOf[nodes.Expression])
```
ml86: Why are we sometimes using `_argumentOut` and sometimes `argumentOut`?
We now see to have all of the xxxIN/OUT method with and without underliner. Whats the reasoning?

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/types/structure/TypeDecl.scala
mpollmeier: really? ok. removed curlies from the `if` block now as well since it's very short

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/utils/ExpandTo.scala
mpollmeier: The difference between them is that `_xxxIn|Out` are defined on `StoredNode`, which means that every node has in/out accessors for *every* edge, even if that's not a legal edge at that position. 
Since we statically know which nodes allow which edges (and also know the neighbor node type) I wanted to replace those with more specific accessors, i.e. define them on the node type (rather than StoredNode), and narrow the type as much as possible. Example: 
```scala
class MethodReturn {
  override def _astIn(): JIterator[StoredNode] = ...
  def astIn: JIterator[Method] = ...
}
```

We had a discussion between us and @bbrehm on slack about this, and came to the conclusion that while it makes sense to add the more specific accessors (more type safety, less casts, help in IDE), @bbrehm pointed out there are still some places (like this one) where we currently don't get around using the 'generic' `_xxxIn|Out` accessors. 
So I'm using the specific accessors wherever possible for now.

dataflowengine/src/main/scala/io/shiftleft/dataflowengine/language/TrackingPoint.scala
mpollmeier: i first did that and was then too afraid since it's a little more than a refactoring. 
but now I have someone else to blame if it goes wrong, thank you :)

PR #617, opened 2020-03-02
reenable travis cache

mpollmeier:


mpollmeier:
ok so all cached dirs are back now, the actual fix was manually clearing the caches via the travis web ui

PR #618, opened 2020-03-03
onlyChecked

bbrehm:
We use `nextChecked` in a lot of places where we actually insist on exactly one item existing. This introduces `onlyChecked` to also assert this fact.

Also make more use of the fast neighbor API. I skipped the `Linker` for now to avoid merge conflicts with https://github.com/ShiftLeftSecurity/codepropertygraph/pull/596 .

### DIFF
--- a/codepropertygraph/src/main/scala/io/shiftleft/Implicits.scala
+++ b/codepropertygraph/src/main/scala/io/shiftleft/Implicits.scala
@@ -19,6 +19,14 @@ object Implicits {
       }
     }
 
+    def onlyChecked: T = {
+      if (iterator.hasNext) {
+        val res = iterator.next
+        assert(!iterator.hasNext)
+        res
+      } else { throw new NoSuchElementException() }
+    }
+
     def nextOption: Option[T] = {
       if (iterator.hasNext) {
         Some(iterator.next)
--- a/codepropertygraph/src/main/scala/io/shiftleft/Implicits.scala
+++ b/codepropertygraph/src/main/scala/io/shiftleft/Implicits.scala
@@ -22,7 +22,7 @@ object Implicits {
     def onlyChecked: T = {
       if (iterator.hasNext) {
         val res = iterator.next
-        assert(!iterator.hasNext)
+        assert(!iterator.hasNext, "iterator was expected to have exactly one element, but it actually has more")
         res
       } else { throw new NoSuchElementException() }
     }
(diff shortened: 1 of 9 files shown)


dataflowengine/src/main/scala/io/shiftleft/dataflowengine/passes/reachingdef/ReachingDefPass.scala
```diff
@@ -277,10 +278,9 @@ class DataFlowFrameworkHelper(graph: ScalaGraph) {
   def getOperation(node: nodes.StoredNode): Option[nodes.StoredNode] = {
     node match {
       case identifier: nodes.Identifier =>
-        if (identifier.argumentIn.hasNext) {
-          getOperation(identifier.argumentIn.nextChecked)
-        } else {
-          None
+        identifier.argumentIn.nextOption match {
+          case None       => None
+          case Some(user) => getOperation(user)
```
Ferada:
```suggestion
        identifier.argumentIn.nextOption.map(getOperation)
```

bbrehm: Oops, thanks!

codepropertygraph/src/main/scala/io/shiftleft/Implicits.scala
```diff
@@ -19,6 +19,14 @@ object Implicits {
       }
     }
 
+    def onlyChecked: T = {
+      if (iterator.hasNext) {
+        val res = iterator.next
+        assert(!iterator.hasNext)
```
mpollmeier:
```suggestion
        assert(!iterator.hasNext, "iterator was expected to have exactly one element, but it actually has more")
```

mpollmeier:
`sbt scalafmt` to make the world a better place, and greenify travis

dataflowengine/src/main/scala/io/shiftleft/dataflowengine/passes/reachingdef/ReachingDefPass.scala
mpollmeier: the suggestion was missing the closing `}`, will remove and merge on green

mpollmeier: also, you wanted `flatMap` rather than `map` here

bbrehm:
Sorry for the delay, I didn't merge it before because I was not yet happy with my local test coverage; i.e. I was not yet confident that there are no realistic situations where the new checks trigger. These hypothetical cases were always logical bugs, but they did not error until this PR.

I guess we'll find out now? Worst case we need to revert, debug the offending code, and re-introduce. 

Regardless, just saying that this merge is high-risk.

PR #622, opened 2020-03-05
Add `FILENAME` property to top-level definitions.

Ferada:
To note is that the method stub creator wasn't updated to deal with other default arguments, so I've removed those and replaced everything with named arguments to prevent any mistakes when attributes are changed in the future.


Ferada:
> Please rename FILENAME to FILE_NAME in order to stick to the naming convention.

I wish that was possible, alas we 1. already have `FILENAME`, 2. Scala (or some code) prevents us from having both `HasFilename` and `HasFileName` generated classes.

Ferada:
I've updated the traversals too. Passes existing tests, also when added to `codescience` everything passes over there too (without having modified any of the frontends mind you).

Running `sptests` now to check whether anything changes over there.

Ferada:
@ml86: This is handling `null` values, aka when the frontends haven't written anything at all. However it also considers `""` an invalid value and does not link up nodes to files if they don't have a filename set. I was considering creating a placeholder file node with path `"<unnamed>"` to capture all of the nodes that actually don't have one set, would that be useful at all?

semanticcpg/src/test/scala/io/shiftleft/semanticcpg/language/types/structure/StructureTests.scala
```diff
@@ -25,7 +25,7 @@ class StructureTests extends WordSpec with Matchers {
     }
 
     "should allow traversing to type declarations" in {
-      cpg.file.typeDecl.name.toSet shouldBe Set("foo")
+      cpg.file.typeDecl.name.toSet shouldBe Set("foo", "void", "int")
```
Ferada: The other two types are now available through this traversal due to the empty file `""`.

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/passes/linking/filecompat/FileLinker.scala
```diff
@@ -0,0 +1,64 @@
...
+    val dstGraph = DiffGraph.newBuilder
+
+    val linker = new Linker(cpg)
+
+    val originalFileNameToNode = mutable.Map.empty[String, nodes.StoredNode]
+    val newFileNameToNode = mutable.Map.empty[String, nodes.NewFile]
+
+    var maxFileOrder = -1
+
+    cpg.graph.graph.vertices().asScala.foreach {
```
ml86: Do `cpg.file...` instead of iterating over all vertices this will only iterate over the FILE nodes.

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/passes/linking/filecompat/FileLinker.scala
```diff
@@ -0,0 +1,64 @@
...
+        NodeTypes.NAMESPACE_BLOCK,
+        NodeTypes.TYPE_DECL,
+        NodeTypes.METHOD
+      ),
+      dstNodeLabel = NodeTypes.FILE,
+      edgeType = EdgeTypes.SOURCE_FILE,
+      dstNodeMap = originalFileNameToNode,
+      dstFullNameKey = "FILENAME",
+      dstGraph,
+      Some((srcNode: nodes.StoredNode, dstFullName: String) => {
```
ml86: Please define this method in a separate `def`

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/passes/linking/filecompat/FileLinker.scala
```diff
@@ -0,0 +1,64 @@
...
+  * SOURCE_FILE edges.
+  *
+  * This pass should come after FILENAME properties have been recovered,
+  * that is, after FileNameCompat.
+  */
+class FileLinker(cpg: Cpg) extends CpgPass(cpg) {
+  override def run(): Iterator[DiffGraph] = {
+    val dstGraph = DiffGraph.newBuilder
+
+    val linker = new Linker(cpg)
```
ml86: Instead of instantiation a linker instance please move the `linkToSingle` method to a linker object.

PR #624, opened 2020-03-11
Use custom exception for `nextChecked` and `onlyChecked`.

fabsx00:
Gremlin swallows `NoSuchElementException` and so failures are silent when `nextChecked` or `onlyChecked` raise in queries. I have replaced `NoSuchElementException` with a custom exception so that Gremlin does not touch it.

### DIFF
--- a/codepropertygraph/src/main/scala/io/shiftleft/Implicits.scala
+++ b/codepropertygraph/src/main/scala/io/shiftleft/Implicits.scala
@@ -2,6 +2,8 @@ package io.shiftleft
 
 object Implicits {
 
+  class NoSuchNodeException extends RuntimeException
+
   /**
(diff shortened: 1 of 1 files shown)


codepropertygraph/src/main/scala/io/shiftleft/Implicits.scala
```diff
@@ -2,6 +2,8 @@ package io.shiftleft
 
 object Implicits {
 
+  class NoSuchNodeException extends RuntimeException
+
   /**
     * A wrapper around a Java iterator that throws a proper NoSuchElementException.
```
ml86: Please update comment according to change.

fabsx00: Done

PR #632, opened 2020-03-13
Add parameterOfType step for TYPE nodes.

ml86:
Also changes expressionOfType to work on class hierarchy instead of
labels so that it now really returns all expressions of a type and not
just IDENTIFIER, CALL and LITERAL.

### DIFF
--- a/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/types/structure/Type.scala
+++ b/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/types/structure/Type.scala
@@ -96,14 +96,17 @@ class Type(val wrapped: NodeSteps[nodes.Type]) extends AnyVal {
   def localsOfType: NodeSteps[nodes.Local] =
     new NodeSteps(raw.in(EdgeTypes.EVAL_TYPE).hasLabel(NodeTypes.LOCAL).cast[nodes.Local])
 
-  def expressionOfType: NodeSteps[nodes.Expression] =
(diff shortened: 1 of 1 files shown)


semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/types/structure/Type.scala
```diff
@@ -100,6 +100,13 @@ class Type(val wrapped: NodeSteps[nodes.Type]) extends AnyVal {
     new NodeSteps(
       raw
         .in(EdgeTypes.EVAL_TYPE)
-        .hasLabel(NodeTypes.IDENTIFIER, NodeTypes.CALL, NodeTypes.LITERAL)
+        .filterOnEnd(_.isInstanceOf[nodes.Expression])
         .cast[nodes.Expression])
+
+  def parameterOfType: NodeSteps[nodes.MethodParameterIn] =
```
fabsx00: Concerning the naming convention: we do not have any other Steps named `of$nodeType`. Is it possible to just make it `parameter` or does that clash?

ml86: I was using the same schema as for `expressionOfType`. But that also seems to be a deviation from the schema. I will change that.

fabsx00: Ok. Approving now so you can merge when changed.

PR #641, opened 2020-03-16
CfgDominator: replace tinkerpop api

mpollmeier:
This defines in a base type of StoredNode for all cfgdominator classes.

### DIFF
--- a/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/passes/cfgdominator/CfgDominator.scala
+++ b/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/passes/cfgdominator/CfgDominator.scala
@@ -1,6 +1,8 @@
 package io.shiftleft.semanticcpg.passes.cfgdominator
 
-class CfgDominator[Node](adapter: CfgAdapter[Node]) {
+import io.shiftleft.codepropertygraph.generated.nodes.StoredNode
+
+class CfgDominator[Node <: StoredNode](adapter: CfgAdapter[Node]) {
 
   /**
     * Calculates the immediate dominators of all CFG nodes reachable from cfgEntry.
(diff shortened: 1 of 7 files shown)


mpollmeier:
Semi-related: since we never use any other type than that base type, we may as well remove the type parameter - wdyt @ml86?

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/passes/cfgdominator/CfgDominator.scala
```diff
@@ -1,6 +1,8 @@
 package io.shiftleft.semanticcpg.passes.cfgdominator
 
-class CfgDominator[Node](adapter: CfgAdapter[Node]) {
+import io.shiftleft.codepropertygraph.generated.nodes.StoredNode
+
+class CfgDominator[Node <: StoredNode](adapter: CfgAdapter[Node]) {
```
ml86: Having the CfgDominator class completely independent of our node representation was a deliberate decision. Why do you want to enforce unnecessary type restrictions here?

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/passes/cfgdominator/CfgDominatorFrontier.scala
```diff
@@ -1,8 +1,9 @@
 package io.shiftleft.semanticcpg.passes.cfgdominator
 
+import io.shiftleft.codepropertygraph.generated.nodes.StoredNode
 import scala.collection.mutable
 
-class CfgDominatorFrontier[Node](cfgAdapter: CfgAdapter[Node], domTreeAdapter: DomTreeAdapter[Node]) {
+class CfgDominatorFrontier[Node <: StoredNode](cfgAdapter: CfgAdapter[Node], domTreeAdapter: DomTreeAdapter[Node]) {
```
ml86: Same as above.

PR #644, opened 2020-03-17
Replace `assert` with `log.error` in `onlyChecked`.

fabsx00:
As discussed over the weekend and in yesterday's call, components that process the CPG shall be a bit more liberal about slight deviations from the spec, and when possible, return results even for slightly broken CPGs, ensuring that these results are in themselves consistent so as to not cause crashes further downstream.

This PR: replacing `assert` with `log.error` in `onlyChecked` for two reasons:
1. if assertions do survive compilation, then they lead to a hard stop, which is not desired
2. whether or not assertions show up depend on the compiler settings, and I'm worried we won't even see them during development.

Overall, I'd propose to stay away from assertions and use the logger instead. In particular, utility methods such as `onlyChecked` should not decide to terminate the program.

### DIFF
--- a/codepropertygraph/src/main/scala/io/shiftleft/Implicits.scala
+++ b/codepropertygraph/src/main/scala/io/shiftleft/Implicits.scala
@@ -1,6 +1,10 @@
 package io.shiftleft
 
+import org.apache.logging.log4j.LogManager
+
 object Implicits {
+
+  private val logger = LogManager.getLogger(classOf[Implicits])
 
   /**
     * A wrapper around a Java iterator that throws a proper NoSuchElementException.
@@ -22,7 +26,9 @@ object Implicits {
     def onlyChecked: T = {
       if (iterator.hasNext) {
         val res = iterator.next
-        assert(!iterator.hasNext, "iterator was expected to have exactly one element, but it actually has more")
+        if (iterator.hasNext) {
+          logger.error("iterator was expected to have exactly one element, but it actually has more")
+        }
         res
       } else { throw new NoSuchElementException() }
     }
@@ -37,3 +43,5 @@ object Implicits {
   }
 
 }
+
+class Implicits {}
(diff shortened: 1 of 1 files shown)


codepropertygraph/src/main/scala/io/shiftleft/Implicits.scala
```diff
@@ -1,7 +1,11 @@
 package io.shiftleft
 
+import org.apache.logging.log4j.LogManager
+
 object Implicits {
 
+  private val logger = LogManager.getLogger(classOf[Implicits])
```
hubertp: Objects are not types but it is not necessary to introduce a companion class just  for the purpose of the logger
```suggestion
  private val logger = LogManager.getLogger(getClass)
```

codepropertygraph/src/main/scala/io/shiftleft/Implicits.scala
```diff
@@ -37,3 +43,5 @@ object Implicits {
   }
 
 }
+
+class Implicits {}
```
hubertp: Not needed. See above comment

bbrehm:
Afaiu assertion failure is just an ordinary exception, just like the `NoSuchElement`-style exceptions we throw otherwise. This is not a panic and should not pull down the process. But it is an error condition: From this moment, we know that all analyses depending on this `onlyChecked()` are incorrect -- i.e. everything in the surrounding try-catch block.

I'm fine with the change, but I think we're better off relying on the already existing try-catch-log around major operations (e.g. loader passes, dataflow queries) than shot-gun logging. A comparison could be made to the `NoSuchElement` exception: We sure could return an element and log the condition instead, but that would be sully.

fabsx00:
I do not see any exception handling for this in callers. Just like the NoSuchElement exception, this is unhandled, and it effectively means that when this assertion is triggered, somebody needs to get up in the middle of the night - just because the input CPG had one edge too many. I would prefer if we report this invalid CPG via our logging infrastructure and look into it the next day. That's what this change does.

PR #655, opened 2020-03-31
Do not write `overlays` field automatically

fabsx00:
Previously, we were automatically writing the names of passes into the `overlays` when these passes finished. That's actually not what we want because an overlay is created by running several passes. Thinking this through and hitting some bugs, I've decided to make use of this field in Ocular and manage applied overlays there.


codepropertygraph/src/main/scala/io/shiftleft/passes/CpgPass.scala
```diff
@@ -33,9 +31,17 @@ import scala.concurrent.duration.DurationLong
...
   import CpgPass.logger
 
+  private val outputName = {
+    if (outName.isEmpty) {
+      this.getClass.getSimpleName()
+    } else {
+      outName
+    }
+  }
+
```
Ferada: Why's that better than before?

fabsx00: Because before, it always returned `Predef$` for some reason...

Ferada: I mean that would make a good comment ...

mpollmeier: > I mean that would make a good comment ...

yes it would, just found myself scratching my head about this one :)

PR #657, opened 2020-04-03
Route all call site to method expansions via ICallResolver interface.

ml86:
This enables us to not write call resolution results directly into the
graph. Instead those results will be provided by the new ICallResolver
method getResolvedCalledMethods and getResolvedCallsites.

From now on all call site to method expansion HAVE TO go via ICallResolver.

### DIFF
--- a/dataflowengine/src/main/scala/io/shiftleft/dataflowengine/passes/reachingdef/ReachingDefPass.scala
+++ b/dataflowengine/src/main/scala/io/shiftleft/dataflowengine/passes/reachingdef/ReachingDefPass.scala
@@ -209,8 +209,8 @@ class ReachingDefPass(cpg: Cpg) extends CpgPass(cpg) {
 class DataFlowFrameworkHelper(graph: ScalaGraph) {
 
   private def callToMethodParamOut(call: nodes.StoredNode): Iterable[nodes.StoredNode] = {
-    ExpandTo
-      .callToCalledMethod(call.asInstanceOf[nodes.Call], NoResolve)
+    NoResolve
+      .getCalledMethods(call.asInstanceOf[nodes.Call])
       .flatMap(method => ExpandTo.methodToOutParameters(method))
   }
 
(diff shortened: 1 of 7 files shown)


dataflowengine/src/main/scala/io/shiftleft/dataflowengine/passes/reachingdef/ReachingDefPass.scala
```diff
@@ -208,12 +208,15 @@ class ReachingDefPass(cpg: Cpg) extends CpgPass(cpg) {
...
+    ExpandTo
+      .callToCalledMethod(call.asInstanceOf[nodes.Call], NoResolve)
+      .flatMap(method => ExpandTo.methodToOutParameters(method))
   }
 
-  private def filterArgumentIndex(vertexList: List[nodes.StoredNode], orderSeq: Seq[Int]): List[nodes.StoredNode] = {
-    vertexList.filter(v => orderSeq.contains(v.asInstanceOf[nodes.HasArgumentIndex].argumentIndex.toInt))
+  private def filterArgumentIndex(vertexList: List[nodes.StoredNode],
+                                  orderSeq: Iterable[Int]): List[nodes.StoredNode] = {
+    vertexList.filter(v => orderSeq.exists(_ == v.asInstanceOf[nodes.HasArgumentIndex].argumentIndex.toInt))
```
fabsx00, line 219: `filter(... _.isInstanceOf[])` or similar to avoid potential class cast exception.

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/ICallResolver.scala
```diff
@@ -1,14 +1,45 @@
...
+
+import scala.collection.mutable
+import scala.jdk.CollectionConverters._
 
 trait ICallResolver {
   def resolveDynamicCallSite(callsite: nodes.Call): Unit
   def resolveDynamicMethodCallSites(method: nodes.Method): Unit
+  def getResolvedCalledMethods(callsite: nodes.CallRepr): Iterable[nodes.Method]
+  def getResolvedCallsites(method: nodes.Method): Iterable[nodes.CallRepr]
+
```
bbrehm, line 15: We need to decide, document and consolidate the API here:

What triggers a resolve? How can I query side-effect-free the status of the resolution? What status can the resolution be in?

I think resolution can be in status: Not started, started, finished. 

95% of the time, people don't care: They want to "resolve this and don't leak implementation internals", i.e. they don't care about the cache.

As such, I propose to remove `resolveDynamicCallSite` and replace it with `getResolvedCalledMethods` that starts a resolution (returns future/callback or list of results), plus `getResolvedCalledMethodsStatus` that immediately returns the resolution status, without triggering any side-effects.

ml86: I pushed an adjust version.

PR #669, opened 2020-04-15
reduce log-level for dynamic call linker fails 

bbrehm:
(these are expected for non-vtable based languages)

### DIFF
--- a/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/passes/linking/calllinker/CallLinker.scala
+++ b/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/passes/linking/calllinker/CallLinker.scala
@@ -13,7 +13,9 @@ import scala.collection.mutable
 import scala.jdk.CollectionConverters._
 
 class CallLinker(cpg: Cpg) extends CpgPass(cpg) {
+
   import CallLinker._
+
   private val methodFullNameToNode = mutable.Map.empty[String, nodes.StoredNode]
 
   /**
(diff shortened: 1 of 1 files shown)


semanticcpg/src/main/scala/io/shiftleft/semanticcpg/passes/linking/calllinker/CallLinker.scala
```diff
@@ -45,7 +45,13 @@ class CallLinker(cpg: Cpg) extends CpgPass(cpg) {
   private def linkCall(call: nodes.Call, dstGraph: DiffGraph.Builder): Unit = {
     val resolvedMethodOption =
       if (call.dispatchType == DispatchTypes.STATIC_DISPATCH) {
-        methodFullNameToNode.get(call.methodFullName)
+        val res = methodFullNameToNode.get(call.methodFullName)
+        if (res == None) {
```
ml86, line 49: Please use `.isEmpty` instead of == None

ml86: I please name the variable appropriately.

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/passes/linking/calllinker/CallLinker.scala
```diff
@@ -61,13 +67,19 @@ class CallLinker(cpg: Cpg) extends CpgPass(cpg) {
                 ._refOut
                 .onlyChecked
 
-              receiverTypeDecl._bindsOut.asScala.collectFirst {
+              val res = receiverTypeDecl._bindsOut.asScala.collectFirst {
                 case binding: nodes.Binding if binding.name == call.name && binding.signature == call.signature =>
                   binding._refOut.onlyChecked.asInstanceOf[nodes.Method]
               }
+              if (res == None) {
+                logger.debug(
```
ml86, line 75: I would never log in this case.

bbrehm: *shrugs*

The old log message was presumably useful (otherwise it wouldn't have been there, right?), just too loud. Hence make info->debug.

ml86: This case went from being considered a problem to now being considered valid input. So why should we log it?

PR #682, opened 2020-04-20
Fix bug in help overview

fabsx00:
Forgot to pass type parameter to `Helper.overview` in generated code, yielding an almost empty table. This PR fixes that.

### DIFF
--- a/console/src/main/scala/io/shiftleft/console/Help.scala
+++ b/console/src/main/scala/io/shiftleft/console/Help.scala
@@ -65,6 +65,7 @@ object Help {
       }
       .mkString("\n")
 
+    val overview = Help.overview[C](tag)
     s"""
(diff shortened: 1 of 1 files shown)


console/src/main/scala/io/shiftleft/console/Help.scala
```diff
@@ -65,14 +65,15 @@ object Help {
...
+    val overview = Help.overview[C](tag)
     s"""
        | class Helper() {
        |
        | $membersCode
        |
        | def run : String = Help.runLongHelp
        |
-       |  override def toString : String = Help.overview
+       |  override def toString : String = \"\"\"${overview}\"\"\"
```
mpollmeier, line 76: a rare sighting of the need to escape characters in a multiline string :)

PR #683, opened 2020-04-20
.help / .helpVerbose for steps

mpollmeier:
Allow users to type `cpg.xyz.help` and get context-sensitive help, i.e. based on the current element type. 
The idea is that all traversal extensions are annotated with a link to the element type, and specific steps (methods) are annotated with some documentation text. 

.helpVerbose also prints the traversal name and also lists the generic steps (e.g. `.toList`) as well as the node-generic steps (e.g. `.label`) if the element is a node. 

Sample output:
```
Available steps for Method:
____________________________________________________________
 step         | description                                |
===========================================================|
 .body        | Traverse to method body (alias for `block`)|
 .myCustomStep| foo bar                                    |
 .namespace   | Traverse to namespace                      |
```

Verbose:
```
Available steps for Method:
___________________________________________________________________________________________________________________________________
 step         | description                                            | traversal name                                           |
==================================================================================================================================|
 .body        | Traverse to method body (alias for `block`)            | io.shiftleft.semanticcpg.language.types.structure.Method |
 .l           | Shorthand for `toList`                                 | io.shiftleft.semanticcpg.language.NodeSteps              |
 .label       | Traverse to node labels                                | io.shiftleft.semanticcpg.language.NodeSteps              |
 .myCustomStep| foo bar                                                | io.shiftleft.semanticcpg.language.StepsTest$MyMethodExt$1|
 .namespace   | Traverse to namespace                                  | io.shiftleft.semanticcpg.language.types.structure.Method |
 .toList      | Execute the traversal and convert the result to a list.| io.shiftleft.semanticcpg.language.NodeSteps              |
```

n.b. earlier commits in this branch experiment with a typeclass-based approach

n.b. I'm using to use java annotations here because I'm using org.reflections to e.g. easily find all classes annotated with a given annotation. That util doesn't support scala annotations, and Scala reflection doesn't allow to give us the same functionality, because it's based on java classloaders, which seem unable to deliver e.g. a simple list of 'all classes available'.

### DIFF
--- a/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/Steps.scala
+++ b/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/Steps.scala
@@ -5,10 +5,12 @@ import io.shiftleft.codepropertygraph.generated.nodes
 import io.shiftleft.codepropertygraph.generated.nodes._
 import java.util.{List => JList}
 
+import io.shiftleft.semanticcpg.{Doc, Doc2, StepsExtJ}
 import org.apache.tinkerpop.gremlin.process.traversal.Scope
 import org.json4s.CustomSerializer
 import org.json4s.native.Serialization.{write, writePretty}
 import org.json4s.Extraction
+import org.reflections.Reflections
 
 import scala.jdk.CollectionConverters._
 import scala.collection.mutable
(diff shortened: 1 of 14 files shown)


semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/Steps.scala
```diff
@@ -86,6 +88,18 @@ class Steps[A](val raw: GremlinScala[A]) {
...
+  /**
+    * Print help/documentation based on the current elementType `A`.
+    * Relies on all step extensions being annotated with @Traversal / @Doc
+    * Note that this works independently of tab completion and implicit conversions in scope - it will simply list
+    * all documented steps in the classpath
+    * */
+  def help()(implicit elementType: ClassTag[A]): String =
+    Help.renderTable(elementType.runtimeClass, verbose = false)
+
+  def helpVerbose()(implicit elementType: ClassTag[A]): String =
```
tuxology, line 100: Is this necessary and is this public? For now, I don't foresee users needing to see this when they hit `help<TAB>` and wanting to individually see traversal names. Could we maybe hide this in `.help.verbose` ?

mpollmeier: it's not 'necessary', but IMO it's useful to the user. 
no strong feelings though, we can drop it, wdyt @fabsx00 ?
alternative would be: `def help(verbose: Boolean = false)`

fabsx00:
Brief conversation with @tuxology - we'll merge for now and start playing with it to see how it feels to the user in practice.

PR #687, opened 2020-04-22
use latest odb-codgen

mpollmeier:
depends on https://github.com/ShiftLeftSecurity/overflowdb-codegen/pull/4

before merging, double check locally if this fixes the problem in https://github.com/ShiftLeftSecurity/ocular/tree/fabs/undobug
n.b. can't do that currently because ocular-latest is red


mpollmeier:
tested, works. can be merged once odb-codegen is (auto-) released. i just merged that pr

PR #691, opened 2020-04-22
rewrite BindingMethodOverridesPass because it was wrong

bbrehm:
It used to mark methods as overridden even if they were not actually overriden. Also, the test coverage was empty.

### DIFF
--- a/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/passes/BindingMethodOverridesPass.scala
+++ b/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/passes/BindingMethodOverridesPass.scala
@@ -38,11 +38,12 @@ class BindingMethodOverridesPass(cpg: Cpg) extends CpgPass(cpg) {
                                 key = NodeKeyNames.IS_METHOD_NEVER_OVERRIDDEN,
                                 value = (!overwritten.contains(binding)).asInstanceOf[AnyRef])
     }
-    return Some(diffGraph.build()).iterator
+    return Iterator(diffGraph.build())
   }
 
   def markRecurse(binding: nodes.Binding): Unit = {
-    if (overwritten.add(binding)) {
+    val wasAlreadyOverwritten = overwritten.add(binding)
+    if (wasAlreadyOverwritten) {
       for (parentType <- binding.bindsIn.next.inheritsFromOut.asScala;
            parentTypeDecl <- parentType.refOut.asScala;
            parentBinding <- bindingTable.get((binding.name, binding.signature, parentTypeDecl))) {
--- a/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/passes/BindingMethodOverridesPass.scala
+++ b/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/passes/BindingMethodOverridesPass.scala
@@ -1,48 +1,53 @@
 package io.shiftleft.semanticcpg.passes
 
 import io.shiftleft.codepropertygraph.Cpg
-import io.shiftleft.codepropertygraph.generated.{NodeKeys, nodes}
-import io.shiftleft.passes.{CpgPass, DiffGraph, ParallelIteratorExecutor}
+import io.shiftleft.codepropertygraph.generated.{nodes, NodeKeyNames}
+import io.shiftleft.passes.{CpgPass, DiffGraph}
 import io.shiftleft.semanticcpg.language._
-import org.apache.logging.log4j.LogManager
+import scala.jdk.CollectionConverters._
 
 class BindingMethodOverridesPass(cpg: Cpg) extends CpgPass(cpg) {
-  import BindingMethodOverridesPass._
+  val overwritten = scala.collection.mutable.HashSet[nodes.Binding]()
+  val bindingTable = scala.collection.mutable.HashMap[(String, String, nodes.TypeDecl), nodes.Binding]()
 
   override def run(): Iterator[DiffGraph] = {
-    val methodIterator = cpg.method.toIterator()
-    new ParallelIteratorExecutor(methodIterator).map(processMethod)
-  }
-
-  private def processMethod(method: nodes.Method): DiffGraph = {
-    val diff = DiffGraph.newBuilder
-    method.start.referencingBinding
-      .toIterator()
-      .foreach(binding => {
-        val typeDecl: Option[nodes.TypeDecl] = binding.start.bindingTypeDecl.headOption()
-        if (typeDecl.isDefined) {
-          val neverOverriddenFlag = isMethodNeverOverridden(typeDecl.get, binding.name, binding.signature)
-          diff.addNodeProperty(
-            binding,
-            NodeKeys.IS_METHOD_NEVER_OVERRIDDEN.name,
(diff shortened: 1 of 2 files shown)


semanticcpg/src/main/scala/io/shiftleft/semanticcpg/passes/BindingMethodOverridesPass.scala
```diff
@@ -1,48 +1,53 @@
...
-    diff.build()
+      }
+    }
+    for (typeDecl <- cpg.typeDecl.toIterator;
+         binding <- typeDecl.bindsOut.asScala) {
+      diffGraph.addNodeProperty(node = binding,
+                                key = NodeKeyNames.IS_METHOD_NEVER_OVERRIDDEN,
+                                value = (!overwritten.contains(binding)).asInstanceOf[AnyRef])
+    }
+    return Some(diffGraph.build()).iterator
```
ml86, line 41: Just `Iterator(diffGraph.build())`

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/passes/BindingMethodOverridesPass.scala
```diff
@@ -1,48 +1,53 @@
...
   }
 
-  def isMethodNeverOverridden(typeDecl: nodes.TypeDecl, bindingName: String, bindingSignature: String): Boolean = {
-    typeDecl.start.derivedTypeDeclTransitive.methodBinding
-      .filter(_.nameExact(bindingName).signatureExact(bindingSignature))
-      .headOption
-      .isEmpty
+  def markRecurse(binding: nodes.Binding): Unit = {
+    if (overwritten.add(binding)) {
```
ml86, line 45: Please give the return value of `add` a proper name for better readability.

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/passes/BindingMethodOverridesPass.scala
```diff
@@ -1,48 +1,54 @@
...
-import io.shiftleft.passes.{CpgPass, DiffGraph, ParallelIteratorExecutor}
+import io.shiftleft.codepropertygraph.generated.{nodes, NodeKeyNames}
+import io.shiftleft.passes.{CpgPass, DiffGraph}
 import io.shiftleft.semanticcpg.language._
-import org.apache.logging.log4j.LogManager
+import scala.jdk.CollectionConverters._
 
 class BindingMethodOverridesPass(cpg: Cpg) extends CpgPass(cpg) {
-  import BindingMethodOverridesPass._
+  val overwritten = scala.collection.mutable.HashSet[nodes.Binding]()
```
Ferada, line 10:
```suggestion
  val overwritten = mutable.HashSet[nodes.Binding]()
```

AFAIK that's how we have the imports elsewhere and it's not as long as the full version.

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/passes/BindingMethodOverridesPass.scala
```diff
@@ -1,48 +1,54 @@
...
+    for (typeDecl <- cpg.typeDecl.toIterator) {
+      val parentTypeDecls = typeDecl.inheritsFromOut.asScala.map {
+        _.refOut.next
+      }.toList
+      for (binding <- typeDecl.bindsOut.asScala) {
+        if (!overwritten.contains(binding)) {
+          val method = binding.refOut.next
+          for (parentTypeDecl <- parentTypeDecls) {
+            val parentBinding = bindingTable.get((binding.name, binding.signature, parentTypeDecl))
+            if (parentBinding.isDefined && parentBinding.get.refOut.next != method) {
```
Ferada, line 28:
```suggestion
            parentBinding.foreach { ... }
```
At least that way don't have to spell out both `isDefined` and `get`.

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/passes/BindingMethodOverridesPass.scala
```diff
@@ -1,48 +1,54 @@
...
+    for (typeDecl <- cpg.typeDecl.toIterator;
+         binding <- typeDecl.bindsOut.asScala) {
+      bindingTable.update((binding.name, binding.signature, typeDecl), binding)
+    }
+    for (typeDecl <- cpg.typeDecl.toIterator) {
+      val parentTypeDecls = typeDecl.inheritsFromOut.asScala.map {
+        _.refOut.next
+      }.toList
+      for (binding <- typeDecl.bindsOut.asScala) {
+        if (!overwritten.contains(binding)) {
```
Ferada, line 24: I'd have maybe said `if (overwritten.contains(binding)) continue` just to reduce the nesting level a bit.

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/passes/BindingMethodOverridesPass.scala
```diff
@@ -1,48 +1,54 @@
...
+              markRecurse(parentBinding.get)
+            }
+          }
         }
-      })
-    diff.build()
+      }
+    }
+    for (typeDecl <- cpg.typeDecl.toIterator;
+         binding <- typeDecl.bindsOut.asScala) {
```
Ferada, line 36: Since this loop happened up there already, I guess they can't / shouldn't be merged or cached?

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/passes/BindingMethodOverridesPass.scala
bbrehm: Yes!

I'd love to do this! Unfortunately scala has neither continue nor break, because it is run by fanatics that put misconceived notions of purity before pragmatic needs of users. 

I'll spare you the extended rant about this language. I have given up "elegant scala", and have settled for "would have been barely readable, if not for scalafmt messing up the formatting".

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/passes/BindingMethodOverridesPass.scala
bbrehm: Afaiu `foreach` opens a stackframe / closure. That's more complex and therefore harder to read than a simple conditional branch.

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/passes/BindingMethodOverridesPass.scala
bbrehm: Yeah, you're right: Theoretically, the "Officially Correct Way (TM)" would be to iterate in topologically sorted order, with respect to inheritance relationship; then we could merge some of the loops. But I think this would be too big of a gun / complicated, and there shouldn't be that many bindings / typeDecls.

I assumed that looping ` for (typeDecl <- cpg.typeDecl.toIterator;
         binding <- typeDecl.bindsOut.asScala)` was roughly as fast as caching the looped-over items.

PR #694, opened 2020-04-23
Print message suggesting `save` after `runAnalyzer`

fabsx00:


console/src/main/scala/io/shiftleft/console/Console.scala
```diff
@@ -324,6 +324,7 @@ abstract class Console[T <: Project](executor: AmmoniteExecutor, loader: Workspa
         serializedCpg.close()
       }
     }
+    report("You may want to `save`")
```
tuxology, line 327:
```suggestion
    report("The graph has been modified. You may want to use the `save` command to save changes on disk. Ocular will also auto-save all the changes to graph collectively when you exit" )
```

PR #709, opened 2020-04-29
Handle block nodes in cfg node expansion.

ml86:


bbrehm:
Can you explain this change? Why and what?

Who needs `lastExpressionInBlock`? Isn't new and old `TrackingPointToCfgNode.apply` doing exactly the same?

ml86:
The lastExpressionInBlock method was moved here from CS repo because it is now also required for CFG expansion. The CFG expansion of a tracking point BLOCK was so far missing which is the cause we dont find JS data flows into sinks like:  foo({sinkArgument})

PR #711, opened 2020-05-01
More docs for `NodeSteps`

fabsx00:


fabsx00:
In general, I agree, however, the scaladocs we had here didn't contain significant information beyond what the method name already encoded, so, I'd rather remove them.

mpollmeier:
> the scaladocs we had here didn't contain significant information beyond what the method name already encoded, so, I'd rather remove them.

true, ok makes sense

PR #713, opened 2020-05-02
refactor schema: nest inNode.name to allow for adding cardinality later

mpollmeier:
the resulting generated nodes are identical

depends on https://github.com/ShiftLeftSecurity/overflowdb-codegen/pull/7
re https://github.com/ShiftLeftSecurity/codepropertygraph/pull/567

### DIFF
--- a/codepropertygraph/src/main/resources/schemas/base.json
+++ b/codepropertygraph/src/main/resources/schemas/base.json
@@ -87,10 +87,29 @@
 		  "LINE_NUMBER", "COLUMN_NUMBER", "LINE_NUMBER_END", "COLUMN_NUMBER_END", "ORDER", "FILENAME"],
          "comment" : "A method/function/procedure",
          "is": ["DECLARATION", "CFG_NODE", "AST_NODE"],
-         "outEdges" : [
-             {"edgeName": "AST", "inNodes": ["METHOD_RETURN", "METHOD_PARAMETER_IN",
-                                             "MODIFIER", "BLOCK", "TYPE_PARAMETER"]},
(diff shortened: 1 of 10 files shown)


fabsx00:
In general, yes, the more we can spec the graph, the better. However, let's only make use of this information in the CPG validator for now and not warn/raise at runtime if cardinality is not correct. I'm currently polishing for a release and this I want to avoid that such a fundamental change causes unforeseen trouble in the release. So, let' merge this Tuesday.

mpollmeier:
Sure, will wait until after the release, this needs to be completed for all json files anyway first. I just wanted to get the 'ok' before doing the repetitive work. 

Fyi as a first step I'm taking the rules from the cpg validator as they are, and simply bake them into the schema. That's mostly so that we know the cardinality when we generate the neighbor accessors. That will make traversing the graph easier and safer. The validators won't change for now.

ml86:
@mpollmeier What you are proposing is less expressive than the current checks in the cpgValidator. E.g. it is not possible to express that a node needs to have a node needs to have a certain incoming edge from A or B. 
I just want to point that out because it will prevent us from doing the incoming/outgoing edge checks in the validator purely based on the data from the JSON files which would have been ideal. Maybe you can change the validator to work with the new represenation in JSON and only the conditions we so far cannot expression in JSON stay in the separate notation.

mpollmeier:
Right, the proposed format cannot express everything we need for the validator. 
There's some copy/pasta in your example, but you're probably referring to e.g. this rule from the OutFactsImporter: `NodeTypes.LITERAL has 1 to N outgoing EdgeTypes.CFG to SuperTypes.Expression or NodeTypes.METHOD_RETURN`. 

For now I'd like to keep that aside, in part to keep the change minimal and the json simple. 
Thanks for bringing that up though, it's good to have that in mind.

PR #734, opened 2020-05-13
Add CPG enhancements for LLVM

AlexDenisov:
I just copied the C part as is. This makes joern load LLVM cpg without issues.
Open questions:
 - what else needs to be adjusted?
 - shall there be more or fewer enhancements for the LLVM frontend?

Anyhow, let this PR be a starting point.


fabsx00:
We need to find out what to do with PolicyLinker, but let's merge this for now.

PR #737, opened 2020-05-14
use facts from validation in schema 1/2

mpollmeier:
part 1/2 is driven by https://github.com/ShiftLeftSecurity/codepropertygraph/blob/master/cpgvalidator/src/main/scala/io/shiftleft/cpgvalidator/facts/OutFactsImporter.scala

each cardinality is handled in a separate commit - only a few required code changes

### DIFF
--- a/codepropertygraph/src/main/resources/schemas/base.json
+++ b/codepropertygraph/src/main/resources/schemas/base.json
@@ -91,7 +91,7 @@
               {"name": "METHOD_RETURN", "cardinality":"1:1"},
               {"name": "METHOD_PARAMETER_IN"},
               {"name": "MODIFIER"},
-              {"name": "BLOCK"},
+              {"name": "BLOCK", "cardinality":"1:1"},
               {"name": "TYPE_PARAMETER"}
             ]
            },
--- a/codepropertygraph/src/main/resources/schemas/base.json
+++ b/codepropertygraph/src/main/resources/schemas/base.json
@@ -313,7 +313,9 @@
                {"name": "UNKNOWN"},
                {"name": "CONTROL_STRUCTURE"}
              ]},
-             {"edgeName": "CFG", "inNodes": [{"name": "METHOD_RETURN"}]},
+             {"edgeName": "CFG", "inNodes": [
+               {"name": "METHOD_RETURN", "cardinality": "0-1:1"}
+              ]},
              {"edgeName": "ARGUMENT", "inNodes": [
                {"name": "CALL"},
                {"name": "IDENTIFIER"},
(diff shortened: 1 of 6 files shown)


codepropertygraph/src/main/resources/schemas/base.json
```diff
@@ -102,7 +102,7 @@
               {"name": "FIELD_IDENTIFIER"},
               {"name": "LITERAL"},
               {"name": "METHOD_REF"},
-              {"name": "METHOD_RETURN"},
+              {"name": "METHOD_RETURN", "cardinality":"n:0-1"},
```
ml86, line 105: This should be 0-1:0-1

codepropertygraph/src/main/resources/schemas/base.json
```diff
@@ -287,7 +289,7 @@
           "is": ["EXPRESSION"],
           "outEdges" : [
             {"edgeName": "CFG", "inNodes": [
-             {"name": "CALL"},
+             {"name": "CALL", "cardinality": "n:1"},
```
ml86, line 292: Should be 0-1:1

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/types/structure/MethodReturn.scala
```diff
@@ -33,7 +33,5 @@ class MethodReturn(val wrapped: NodeSteps[nodes.MethodReturn]) extends AnyVal {
     new NodeSteps(raw.out(EdgeTypes.EVAL_TYPE).cast[nodes.Type])
 
   def toReturn: NodeSteps[nodes.Return] =
-    new NodeSteps(raw.flatMap { mr =>
-      __(mr.toReturn.toSeq: _*)
-    })
+    new NodeSteps(raw.map(_.toReturn).collect { case Some(ret) => ret })
```
ml86, line 36: Just `.flatMap(_.toReturn)` should be enough.

mpollmeier: unfortunately that doesn't compile, because `_.toReturn` returns an `Option[nodes.Return]`, and `flatMap` needs a `Steps` instance. 

Btw once we swap to Odb Traversal the flatMap you proposed should work (since Odb Traversal is a scala collection): 
```
List(1,2,3).flatMap{ _ match { case 1|2 => None; case 3 => Some(4)} }
val res2: List[Int] = List(4)
```

codepropertygraph/src/main/resources/schemas/base.json
mpollmeier: done, thanks!

codepropertygraph/src/main/resources/schemas/base.json
mpollmeier: done, thanks!

PR #738, opened 2020-05-14
Modify LayerCreatorAPI to allow passing of options by user

fabsx00:
I have been busy creating a sample extension for Ocular/Joern and noticed that we currently had no easy way of passing options to extensions. This also uncovered that running `run.ossdataflow` in `joern` would fail with an NPE because the semantics could not be passed to the analyzer.

This PR fixes this problem. On start of Ocular/Joern, and `opts` object is created with a field for each dynamically determined extension, e.g., there is now `opts.ossdataflow`. Users can set options on the shell before invoking the extension via `run.$extension`.

### DIFF
--- a/console/src/main/scala/io/shiftleft/console/Run.scala
+++ b/console/src/main/scala/io/shiftleft/console/Run.scala
@@ -20,8 +20,21 @@ object Run {
       .map(t => (t.getSimpleName.toLowerCase, t.getName))
       .filter(t => !exclude.contains(t._2))
 
+    val optsMembersCode = layerCreatorTypeNames
+      .map { case (varName, typeName) => s"val $varName = $typeName.defaultOpts" }
+      .mkString("\n")
+
+    val optsCode =
+      s"""
+        |class OptsDynamic {
+        | $optsMembersCode
+        |}
+        |
+        |val opts = new OptsDynamic()
+        |""".stripMargin
+
     val membersCode = layerCreatorTypeNames
-      .map { case (varName, typeName) => s"def $varName: Cpg = _runAnalyzer(new $typeName())" }
+      .map { case (varName, typeName) => s"def $varName: Cpg = _runAnalyzer(new $typeName({ () => opts.$varName}))" }
       .mkString("\n")
 
     val toStringCode =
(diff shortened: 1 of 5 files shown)


console/src/main/scala/io/shiftleft/console/Run.scala
```diff
@@ -20,8 +20,21 @@ object Run {
...
+        |class OptsDynamic {
+        | $optsMembersCode
+        |}
+        |
+        |val opts = new OptsDynamic()
+        |""".stripMargin
+
     val membersCode = layerCreatorTypeNames
-      .map { case (varName, typeName) => s"def $varName: Cpg = _runAnalyzer(new $typeName())" }
+      .map { case (varName, typeName) => s"def $varName: Cpg = _runAnalyzer(new $typeName({ () => opts.$varName}))" }
```
mpollmeier, line 37: this change seems to cause ocular to fail...
```
./updateDependencies.sh --non-interactive
./testDistro.sh
```

results in 
```
CodePredef.sc:29: object Run$$anon$1 is not a member of package io.shiftleft.console
val run$$anon$1 = io.shiftleft.console.Run$$anon$1.defaultOpts
                                       ^
CodePredef.sc:49: type Run$$anon$1 is not a member of package io.shiftleft.console
def run$$anon$1: Cpg = _runAnalyzer(new io.shiftleft.console.Run$$anon$1(opts.run$$anon$1))
                                                             ^
CodePredef.sc:58: object Run$$anon$1 is not a member of package io.shiftleft.console
   List(List("semanticcpg",io.shiftleft.semanticcpgext.layers.Semanticcpg.description.trim), List("securityprofile",io.shiftleft.securityprofile.layers.Securityprofile.description.trim), List("commit",io.shiftleft.console.Commit.description.trim), List("run$$anon$1",io.shiftleft.console.Run$$anon$1.description.trim), List("dataflow",io.shiftleft.dataflowengine.layers.Dataflow.description.trim), List("tagging",io.shiftleft.securityprofile.layers.Tagging.description.trim))
                                                                                                                                                                                                                                                                                                ^
Exception in thread "main" java.lang.AssertionError: script failed: Compilation Failed
```

PR #741, opened 2020-05-16
Fix `importCode` on very large code bases

fabsx00:
- use `--overflowdb` for fuzzyc by default, avoids serialization to proto and back
- remove expensive query that touches all edges in `Scpg::probe`
- ~~avoid memory spike when closing bin.zip for overlays. Apparently, the inverse proto overlays are not written out until `close` is called, and there is no flush method. I added the next best thing: a method that calls close and reopens the filesystem. Related reading: http://mail.openjdk.java.net/pipermail/nio-dev/2012-July/001763.html~~
- do not use zip to store undo information (serializedCpg) and use simple directories instead. It turned out that ZipFileSystem created large memory spikes on close and the zip operation itself is not free either.

**Update**: let me see if we can actually get around zipping. Since the use-case has changed (we are no longer looking to use the zip file to exchange CPGs across systems but only to bundle undo information), we may be fine not zipping at all and writing out directories instead.

### DIFF
--- a/codepropertygraph/src/main/scala/io/shiftleft/SerializedCpg.scala
+++ b/codepropertygraph/src/main/scala/io/shiftleft/SerializedCpg.scala
@@ -1,21 +1,17 @@
 package io.shiftleft
 
-import java.io.{File, IOException}
-import java.net.{URI, URISyntaxException}
-import java.nio.file.{FileSystem, FileSystems, Files}
-import java.util
-
+import java.io.IOException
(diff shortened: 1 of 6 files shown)


fabsx00:
Glad I took the time to debug this. In short, `importCode` now works on large code bases like firefox. Previously, only `joern-parse` followed by `importCpg` worked. A nice side effect is also that we use much less ram, and therefore, the overflow mechanism kicks in much later.

PR #747, opened 2020-05-18
RIP OverlaySP loader

fabsx00:
Once https://github.com/ShiftLeftSecurity/codescience/pull/3797 is merged, there are no more components that make use of the old "sp.bin.zip" SP and so we can now remove the corresponding loader code to avoid confusion in the future.


fabsx00:
Thinking about it a bit more, let's close this one for now. Who knows if we may still need this in the near future.

PR #757, opened 2020-05-19
Handle old undo info format

fabsx00:
If we detect a zip file instead of a directory of zip files, then pass that to undo.

### DIFF
--- a/console/src/main/scala/io/shiftleft/console/Console.scala
+++ b/console/src/main/scala/io/shiftleft/console/Console.scala
@@ -224,8 +224,13 @@ class Console[T <: Project](executor: AmmoniteExecutor, loader: WorkspaceLoader[
   def undo: File = {
     project.overlayDirs.lastOption
       .map { dir =>
-        val zipFiles = dir.list.toList.sortWith(compareFiles).reverse.map(_.path.toString)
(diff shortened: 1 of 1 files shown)


console/src/main/scala/io/shiftleft/console/Console.scala
```diff
@@ -224,8 +224,13 @@ class Console[T <: Project](executor: AmmoniteExecutor, loader: WorkspaceLoader[
   def undo: File = {
     project.overlayDirs.lastOption
       .map { dir =>
-        val zipFiles = dir.list.toList.sortWith(compareFiles).reverse.map(_.path.toString)
-        CpgLoader.addDiffGraphs(zipFiles, cpg)
+        if (dir.isRegularFile) {
+          System.err.println("Detected undo information in old format. That's ok.")
```
hubertp, line 228: `System.err` doesn't sound "ok"?

fabsx00: We use System.err in Ocular to report things back to the user. We do that because Ocular may also run in interactive scripts and then these messages don't end up in stdout.

hubertp: If that's the norm, then ok. I would really like to see us using some logger here instead.

PR #760, opened 2020-05-20
Pass optimization

fabsx00:
We have a few passes that create one large diff graph for the entire program in a single thread. The DiffGraph overhead is low at this point, but keeping graphs that grow with the size of the input will send us into OOMs (confirmed). Making use of the new ParallelCpgPass class to avoid this. This affects:

- TypeDeclStubCreator
- CallLinker
- CapturingLinker
- FileNameCompat
- MethodDecoratorPass


semanticcpg/src/main/scala/io/shiftleft/semanticcpg/passes/linking/calllinker/CallLinker.scala
```diff
@@ -12,36 +12,37 @@ import org.apache.logging.log4j.{LogManager, Logger}
...
-        case exception: Exception =>
-          throw new RuntimeException(exception)
-      }
-    }
+  override def partIterator: Iterator[nodes.Call] = cpg.call.toIterator()
 
-    Iterator(dstGraph.build())
+  /**
+    * Main method of enhancement - to be implemented by child class
+    **/
```
Ferada, line 35:
```suggestion
```

PR #764, opened 2020-05-22
replace tinkerpop api with neighbor accessors / scala collections

mpollmeier:
* replace tinkerpop sack traversal with standard for loop: Expression.parameter, MethodParameter.argument, MethodParameterOut.argument
* TrackingPointMethodsBase without tinkerpop
* some more easy refactorings away from tinkerpop
* wherever it's easily portable already, don't start traversals on node level

### DIFF
--- a/dataflowengine/src/main/scala/io/shiftleft/dataflowengine/language/Path.scala
+++ b/dataflowengine/src/main/scala/io/shiftleft/dataflowengine/language/Path.scala
@@ -19,7 +19,7 @@ object Path {
 
         val trackedSymbol = trackingPoint match {
           case _: nodes.MethodParameterIn =>
-            val paramsPretty = method.parameter.l.sortBy(_.order).map(_.code).mkString(", ")
+            val paramsPretty = method.parameter.toList.sortBy(_.order).map(_.code).mkString(", ")
             s"$methodName($paramsPretty)"
           case _ => trackingPoint.cfgNode.repr
         }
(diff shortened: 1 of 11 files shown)


dataflowengine/src/main/scala/io/shiftleft/dataflowengine/language/Path.scala
```diff
@@ -19,7 +19,7 @@ object Path {
 
         val trackedSymbol = trackingPoint match {
           case _: nodes.MethodParameterIn =>
-            val paramsPretty = method.parameter.l.sortBy(_.order).map(_.code).mkString(", ")
+            val paramsPretty = method.parameter.toList.sortBy(_.order).map(_.code).mkString(", ")
```
fabsx00, line 22: The `.l` shorthand is pretty handy. Wonder whether we should introduce it on top of scala collections at some point. Not in this PR though.

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/nodemethods/MethodMethods.scala
```diff
@@ -7,17 +7,17 @@ import io.shiftleft.Implicits.JavaIteratorDeco
 
 class MethodMethods(val node: nodes.Method) extends AnyVal {
 
-  def parameter: NodeSteps[nodes.MethodParameterIn] =
-    node.start.parameter
+  def parameter: Iterator[nodes.MethodParameterIn] =
+    node._methodParameterInViaAstOut
```
fabsx00, line 11: I know the type is called `MethodParameterIn`, but we usually just refer to it as `parameter` throughout the rest of the code. Or is this generated?

mpollmeier: `_methodParameterInViaAstOut` is generated

mpollmeier:
> One general note: Gremlin swallows NoSuchElement exceptions, so, whenever code is ported over from Gremlin to scala iterators, we probably want to introduce exception handling to catch, report and ignore NoSuchElement.

It depends on the use case. But yes, this is an important difference when using normal collections vs gremlin.

PR #774, opened 2020-05-27
Allow running custom write queries via `run(query)`

fabsx00:
Currently, the only way of editing the graph was by writing a complete extension, which would then appear in the list of extensions returned by `run`. If one wants to only quickly tag some nodes on the shell via `cpg.method.newTagNode()`, that would not work because a more complex dance is necessary to create the layer and store the undo information. This PR fixes that by offering `run(query)` where query just needs to have a `store` method, e.g., it is an instance of `NewNodeSteps` or of `NewTagNodePair`. Follow-up PR in the documentation is on its way.


tuxology:
Thanks!

PR #784, opened 2020-06-05
Rename `callOut` to `call` to be consistent

fabsx00:
When traversing from a method to all of its `call` nodes, we currently need to use `callOut`. This is inconsistent as traversing to all nodes in the method of type `type` is achieved via the step `.type` in most other cases. Deprecating callOut to make the change backward compatible.


semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/callgraphextension/Method.scala
```diff
@@ -69,6 +69,6 @@ class Method(val wrapped: NodeSteps[nodes.Method]) extends AnyVal {
     * Outgoing call sites to methods where fullName matches `regex`.
     * */
   def callOutRegex(regex: String)(implicit callResolver: ICallResolver): NodeSteps[nodes.Call] =
```
mpollmeier, line 71: apply the same change for `callOutRegex`?

fabsx00: I tried, but that leads to ambiguous implicit conversions :/

PR #796, opened 2020-06-10
use latest added odb graph api

mpollmeier:
depends on
* https://github.com/ShiftLeftSecurity/overflowdb/pull/110
* https://github.com/ShiftLeftSecurity/overflowdb-codegen/pull/14

### DIFF
--- a/codepropertygraph/build.sbt
+++ b/codepropertygraph/build.sbt
@@ -3,7 +3,7 @@ name := "codepropertygraph"
 dependsOn(Projects.protoBindings)
 
 libraryDependencies ++= Seq(
-  "io.shiftleft" %% "overflowdb-traversal" % "0.0.0-MP",
+  "io.shiftleft" %% "overflowdb-traversal" % "0.87",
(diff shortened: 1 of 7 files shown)


mpollmeier:
Hmm, so currently it's the other way around, everywhere. NodeRef is exposed to the user (because it's thin, it only holds the reference to the serializable OdbNode). The nodes we hand around are the NodeRefs, e.g. `nodes.Method` is a NodeRef. `nodes.MethodDB` doesn't appear in any api because it's an `OdbNode`, and that's by design. 

I agree that the naming is not ideal, and the fact that the most basic thing to pass around has a type parameter sucks. Maybe we should introduce a root interface for the two, which we use in APIs? 

Either way, this PR doesn't make any semantic changes (NodeRef has already replaces Vertex elsewhere), so I'd suggest we discuss that in a separate context and I'll create follow-up PRs to avoid piling up commits.

ml86:
I am well aware of the split between NodeRef and the DB nodes. I just do not want the name NodeRef be reflected in the ODB using code. If you need to introduce a new interface for that, thats ok.

PR #797, opened 2020-06-11
Add node type step for control structures

fabsx00:
`cpg.method.controlStructure` was possible before, but we were missing `cpg.controlStructure`.


semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/NodeTypeStarters.scala
```diff
@@ -32,6 +32,9 @@ class NodeTypeStarters(cpg: Cpg) {
     * */
   def comment(code: String): NodeSteps[nodes.Comment] =
     comment.code(code)
+  @Doc("All control structures (source-based frontends)")
```
ursachec, line 35:
```suggestion

@Doc("All control structures (source-based frontends)")
```

PR #803, opened 2020-06-13
Enforce `code` field for CFG nodes

fabsx00:
*This is part of work I am doing on exposing a useful language for CFG traversals, including traversals over the control dependence graph and dominator trees. These provide a more robust tool than AST traversals alone as they can take into account unstructured control flow. We do currently have some CFG-based filtering in conjunction with the data flow tracker, but these rely on data flows to be detected first, and hence, they can only be used when (a) a sufficient amount of policies is available, and (b) for languages where data-flow tracking works well.*

We previously had two methods for raw CFG traversal: `cfgPrev` and `cfgNext`. Contrary to expectation, these were defined to return `Expressions` as opposed for `CfgNodes`. In a previous PR, I moved them to `CfgNodes`. It turns out now that some scripts assume that `CfgNodes` have a `code` field, and to my surprise, we did not enforce this earlier. It turns out that only two CFG nodes are missing this field: JUMP_TARGET (recently introduced and only used in the C frontend so far) and `METHOD`. 

For JUMP_TARGET, I will fill out this field in the C-frontend now. For methods, we can leave it empty for now and gradually place the method header or something similar there on a per-frontend basis. The code field is frontend-dependent anyway, being similar to parser-type-name.

This should also fix `ocular-latest-deps`.

### DIFF
--- a/codepropertygraph/src/main/resources/schemas/base.json
+++ b/codepropertygraph/src/main/resources/schemas/base.json
@@ -81,7 +81,7 @@
         // Nodes of method declarations
 
         {"id" : 1, "name" : "METHOD",
-         "keys": ["NAME", "FULL_NAME", "IS_EXTERNAL", "SIGNATURE", "AST_PARENT_TYPE", "AST_PARENT_FULL_NAME",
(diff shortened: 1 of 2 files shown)


fabsx00:
`JUMP_TARGET` is currently not included in the control flow graph, that's true. It still derives from CfgNode though. I do believe it makes sense to even include it in the CFG in the future so that we can use `cfgNext`/`cfgPrev` on it. Low-priority though.

PR #804, opened 2020-06-13
Add `controlledBy`/`controls` method for CFG traversals

fabsx00:
*This is part of work I am doing on exposing a useful language for CFG traversals, including traversals over the control dependence graph and dominator trees. These provide a more robust tool than AST traversals alone as they can take into account unstructured control flow. We do currently have some CFG-based filtering in conjunction with the data flow tracker, but these rely on data flows to be detected first, and hence, they can only be used when (a) a sufficient amount of policies is available, and (b) for languages where data-flow tracking works well.*

A utility method named `getControllingNodesFromCdg` was available in `io.shiftleft.dataflowengine`: getting controlling nodes from the CDG seems generally useful, and so, this PR adds it to the language as `controlledBy`. I have also added the counterpart `controls`, which walks CDG edhes in the opposite direction. Once this is merged, we can optionally remove the utility method in `io.shiftleft.dataflowengine`.

### DIFF
--- a/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/types/expressions/generalizations/CfgNode.scala
+++ b/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/types/expressions/generalizations/CfgNode.scala
@@ -67,4 +67,7 @@ class CfgNode[A <: nodes.CfgNode](val wrapped: NodeSteps[A]) extends AnyVal {
   def controlledBy: NodeSteps[nodes.CfgNode] =
     wrapped.flatMap(_.controlledBy)
 
+  def controls: NodeSteps[nodes.CfgNode] =
+    wrapped.flatMap(_.controls)
+
 }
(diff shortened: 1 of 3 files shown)


semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/types/expressions/generalizations/CfgNode.scala
```diff
@@ -60,4 +60,14 @@ class CfgNode[A <: nodes.CfgNode](val wrapped: NodeSteps[A]) extends AnyVal {
...
     )
 
+  /**
+    * Recursively determine all nodes on which any of
+    * the nodes in this traversal are control dependent
+    * */
+  def controlledBy: NodeSteps[nodes.CfgNode] =
+    wrapped.flatMap(_.controlledBy)
+
+  def controls: NodeSteps[nodes.CfgNode] =
```
ml86, line 70: Missing function describtion.

PR #805, opened 2020-06-15
adaptations for odb public api changes:

mpollmeier:
* Node -> CpgNode
* NodeRef -> Node
* remove unnecessary type lower bounds


mpollmeier:
> Is it now CpgNode but NewNode?

The only reason for renaming `Node` -> `CpgNode` was to allow to introduce `overflowdb.Node`. Personally I'd keep `NewNode` as is, but if you like we can rename `NewNode` -> `NewCpgNode` - just let me know.

PR #816, opened 2020-06-25
Introduce `changeWorkspace`

fabsx00:
Part of the work on ocular-server.

We were previously limited to a single workspace (located in the installation directory). For Ocular server, we want to be able to offer access to multiple workspaces and allow workers to change the workspace they are working on.

### DIFF
--- a/console/src/main/scala/io/shiftleft/console/Console.scala
+++ b/console/src/main/scala/io/shiftleft/console/Console.scala
@@ -26,7 +26,7 @@ class Console[T <: Project](executor: AmmoniteExecutor, loader: WorkspaceLoader[
   def console: Console[T] = this
 
   protected var workspaceManager: WorkspaceManager[T] = _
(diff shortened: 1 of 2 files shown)


console/src/main/scala/io/shiftleft/console/Console.scala
```diff
@@ -25,8 +25,10 @@ class Console[T <: Project](executor: AmmoniteExecutor, loader: WorkspaceLoader[
   def config: ConsoleConfig = _config
   def console: Console[T] = this
 
-  protected val workspacePathName: String = config.install.rootPath.path.resolve("workspace").toString
-  protected val workspaceManager = new WorkspaceManager[T](workspacePathName, loader)
+  protected var workspaceManager: WorkspaceManager[T] = _
+  changeWorkspace(config.install.rootPath.path.resolve("workspace").toString)
```
prabhu, line 29: switchWorkspace?

PR #818, opened 2020-06-25
freshen up `id` steps: we always use `Long` ids

mpollmeier:
queries like `cpg.method.id(1L).l` compile now - thanks @ursachec for reporting!


mpollmeier:
`id2` is temporary, during the transition away from tinkerpop, which defines `id: Object` for all elements, while we want `id: Long`

PR #823, opened 2020-07-02
Embedded Ammonite Shell and Tests

fabsx00:
This PR introduces an embedded ammonite shell in `io.shiftleft.console`. The shell is internally controlled by holding a pipe to ammonite's stdin and pipes from its stdout and stderr. From the user perspective, queries can be submitted to the shell and results are returned (synchronous mode), however, it is also possible to submit a query and invoke a method once the result is ready (asynchronous mode). Please see the unit tests for usage examples.

**Assumptions:**
- We assume that a single embedded Ammonite shell services only a single user at once. Anything else seems to ask for trouble as it would be equivalent to multiple users typing on the same terminal - interleaving commands while the shell holds only a single state.
- We do however want to allow this user to post a sequence of commands, e.g., via HTTP, and then process results as they become available.

**Design**
- Ammonite is started in a thread and we keep pipes to stdin and from stdout and stderr to communicate with it
- A second thread - the UserThread - holds these pipes and blocks on a queue.
- Upon calls to `.queryAsync`, a job is placed in the queue, which the UserThread then fetches and services - performing the necessary communication with Ammonite via pipes. Upon completion, it calls the `observer` passed to `.queryAsync`
- For convenience, we also offer `query`, which blocks until `queryAsync` has completed, then fetches the result and returns it to the caller.

**Some things you may find interesting**
- With the default ammonite frontend, an embedded shell cannot be run in environments where `/dev/tty` is not properly configured, e.g., in certain containers and in IntelliJ tests. The solution is to provide a custom ammonite frontend, which we do via predef
- Ammonite does not allow disabling terminal echo, however, we can disable it by replacing the `displayTransform` in our custom frontend with a function that swallows its input
- With the default frontend and redirected stdin, ctrl+c is not received on the shell. This problem can be fixed by spawning a separate thread that listens on stdin and then sends SIGINT to ammonite, a route I initially followed. Upon replacing the default frontend, however, this problem no longer existed.

### DIFF
--- a/console/src/main/scala/io/shiftleft/console/embammonite/EmbeddedAmmonite.scala
+++ b/console/src/main/scala/io/shiftleft/console/embammonite/EmbeddedAmmonite.scala
@@ -1,0 +1,154 @@
+package io.shiftleft.console.embammonite
+
+import java.io.{BufferedReader, InputStreamReader, PipedInputStream, PipedOutputStream, PrintWriter}
+import java.util.UUID
+import java.util.concurrent.{BlockingQueue, LinkedBlockingQueue, Semaphore}
+
+import ammonite.util.Colors
+import org.apache.logging.log4j.LogManager
+
+/**
+  * Result of executing a query, containing in particular
+  * output received on standard out and on standard error.
+  * */
+class QueryResult(val out: String, val err: String)
+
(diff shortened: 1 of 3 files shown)


itsacoderepo:
:+1:

PR #828, opened 2020-07-06
Ocular/Joern WebSocket Server

fabsx00:
A simple HTTP server that allows an Ocular/Joern instance to be remote controlled via HTTP. As queries can take a while to complete, we allow posting queries and coming back for results later. To ensure that the client does not need to poll, we use websockets to inform the client whenever new results are ready. The routes we offer are the following:

* POST: /query : enqueue a query. Returns a UUID
* GET: /result/:uuid : retrieve result with given uuid.
* WS: /connect: subscribe to the WebSocket to receive notifications for query completions. A notification simply consists of the query UUID, making it possible to match queries to reponses.

### DIFF
--- a/console/src/main/scala/io/shiftleft/console/wsserver/WebsocketServer.scala
+++ b/console/src/main/scala/io/shiftleft/console/wsserver/WebsocketServer.scala
@@ -1,16 +1,30 @@
 package io.shiftleft.console.wsserver
 
+import ujson.Obj
+
 class WebsocketServer extends cask.MainRoutes {
 
-  @cask.websocket("/connect/:userName")
-  def handler(userName: String): cask.WebsocketResult = {
-    cask.WsHandler { channel =>
-      cask.WsActor {
-        case cask.Ws.Text("") => channel.send(cask.Ws.Close())
-        case cask.Ws.Text(data) =>
-          channel.send(cask.Ws.Text(userName + " " + data))
-      }
+  var openConnections = Set.empty[cask.WsChannelActor]
+
+  @cask.websocket("/connect")
+  def handler(): cask.WebsocketResult = {
+    cask.WsHandler { connection =>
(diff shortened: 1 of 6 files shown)


console/src/test/scala/io/shiftleft/console/wsserver/WebSocketServerTests.scala
```diff
@@ -0,0 +1,78 @@
...
+
+    "allow connecting, posting a query, and fetching result" in Fixture() { host =>
+      var wsPromise = scala.concurrent.Promise[String]
+      cask.util.WsClient.connect(s"$host/connect") {
+        case cask.Ws.Text(msg) => wsPromise.success(msg)
+      }
+
+      // Connect to websocket endpoint
+
+      val wsMsg = Await.result(wsPromise.future, Inf)
```
ursachec, line 27: Not sure how the test will behave if the implementation blocks for too long, I'd add a timeout

```suggestion
      val wsMsg = Await.result(wsPromise.future, Duration(10, SECONDS))
```

console/src/main/scala/io/shiftleft/console/wsserver/WebsocketServer.scala
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  @cask.get("/result/:uuidStr")
+  def getResult(uuidStr: String): Obj = {
+    val uuid = UUID.fromString(uuidStr)
+    val result = resultMap.remove(uuid)
+    if (result == null) {
+      ujson.Obj("success" -> false)
+    } else {
+      ujson.Obj("success" -> true, "uuid" -> result.uuid.toString, "out" -> result.out, "err" -> result.err)
```
ursachec, line 44: shouldn't this be something like `IF result.err == null THEN return object.success.true ELSE return object.success.false`?

PR #837, opened 2020-07-15
Add basic sketch for cpgqls-client-python

ursachec:


cpgqls-client-python/README.md
```diff
@@ -0,0 +1,55 @@
+## cpgqls-client
+
+`cpgqls-client` is a simple Python library for communicating with an instace of
```
fabsx00, line 3: instace => instance

cpgqls-client-python/LICENSE
```diff
@@ -0,0 +1,213 @@
...
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+
+========================================================================
+MIT Licenses
+========================================================================
+
+The Apache TinkerPop project bundles the following components under the MIT License:
```
fabsx00, line 208: I think we're not using tinkerpop in this lib, so, if we place a separate license in the lib folder, then we should omit references to it.

ursachec: My bad, updated

ursachec:
Discussed IRL, moving the library to a separate repo, will close the PR and point to it as soon as the task is done

ursachec:
Moved to different repository https://github.com/ShiftLeftSecurity/cpgqls-client-python

PR #838, opened 2020-07-15
Use slf4j as the primary logging facade

hubertp:
Switching to slf4j completely. The implementation will then decide what
logging library we will use, based on the dependency included.

Looks like fuzzy2cpg is bring logback as a dependency.
Should be fixed in https://github.com/ShiftLeftSecurity/fuzzyc2cpg/pull/204


hubertp:
@fabsx00 I have like 4-5 follow up PRs in downstream projects in the pipeline that cleanup that spaghetti of different loggers. I will only merge this one once i have all of them lined up

mpollmeier:
:partying_face:

PR #840, opened 2020-07-16
astChilren: order nodes by `order` field

fabsx00:
While `astChildren` already returns nodes in the order they were added, and therefore, coincidentally, ordered by the `order` field, this is a property of the backend storage and it may change in the future. We want to guarantee users of `astChildren` that children are returned in the correct order, and therefore, I am adding an explicit ordering step.


fabsx00:
(Rebuild completed successfully, just not showing up)

PR #842, opened 2020-07-17
Ordering for `astChildren` + node method

fabsx00:
Minor fixes for `astChildren`:
- `astChildren` was previously only defined on steps but not on nodes
- Ensure that siblings are ordered by `order`. Note that it is not enough to simply order the entire pipe, we want to order children per parent.

### DIFF
--- a/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/types/expressions/generalizations/AstNode.scala
+++ b/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/types/expressions/generalizations/AstNode.scala
@@ -45,11 +45,13 @@ class AstNode[A <: nodes.AstNode](val wrapped: NodeSteps[A]) extends AnyVal {
   def astMinusRoot: NodeSteps[nodes.AstNode] =
     new NodeSteps(raw.repeat(_.out(EdgeTypes.AST)).emit.cast[nodes.AstNode])
 
+  import scala.jdk.CollectionConverters._
+
   /**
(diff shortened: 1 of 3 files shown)


semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/types/expressions/generalizations/AstNode.scala
```diff
@@ -45,11 +45,13 @@ class AstNode[A <: nodes.AstNode](val wrapped: NodeSteps[A]) extends AnyVal {
...
 
+  import scala.jdk.CollectionConverters._
+
   /**
-    * Direct children of node in the AST
+    * Direct children of node in the AST. Siblings are ordered by their `order` fields
     * */
   def astChildren: NodeSteps[nodes.AstNode] =
-    new NodeSteps(raw.out(EdgeTypes.AST).cast[nodes.AstNode])
+    new NodeSteps(raw.flatMap(_._astOut().asScala.toList.start.cast[nodes.AstNode].orderBy(_.order).raw))
```
fabsx00, line 54: Tried different versions, all ending up in a wild dance from iterator to scala list to gremlin traversal and domain traversal. These kinds of dances will hopefully resolve themselves when we use ODB traversals natively.

mpollmeier: yup that'll look a lot nicer

PR #851, opened 2020-07-20
Rename WebsocketServer to CPGQLServer

ursachec:


console/src/main/scala/io/shiftleft/console/wsserver/WebsocketServer.scala
```diff
@@ -6,6 +6,10 @@ import java.util.concurrent.ConcurrentHashMap
 import io.shiftleft.console.embammonite.{EmbeddedAmmonite, QueryResult}
 import ujson.Obj
 
+@Deprecated
```
fabsx00, line 9: We just released this on Friday, so, no need to go through deprecation cycle

PR #852, opened 2020-07-20
Make CPGServer port and host configurable

ursachec:


console/src/main/scala/io/shiftleft/console/cpgqlserver/CPGQLServer.scala
```diff
@@ -6,7 +6,15 @@ import java.util.concurrent.ConcurrentHashMap
 import io.shiftleft.console.embammonite.{EmbeddedAmmonite, QueryResult}
 import ujson.Obj
 
-class CPGQLServer(ammonite: EmbeddedAmmonite) extends cask.MainRoutes {
+class CPGQLServer(ammonite: EmbeddedAmmonite, serverHost: String, serverPort: Int) extends cask.MainRoutes {
```
fabsx00, line 9: If you say `ammonite: EmbeddedAmmonite, val serverHost : String, val serverPort : Int)` then you do not need the two getters below.

fabsx00: Might also be a good place for default values, but up to you.

ursachec: @fabsx00 Those aren't setters, they're overwritten methods of `cask.MainRoutes`. Don't suppose there's a better shortcut for them... As to the default values, I'd rather leave them out in this case

ursachec: Actually, could be that you're right, not sure, checking...

ursachec: It seems I was correct, merging as is and cleaning up later if necessary

PR #855, opened 2020-07-21
Allow "GO" for existing Go CPG's.

Ferada:
`GO` is the value being used in `go2cpg`. Happy to change it, although we also want to load the older files that already exist.

Fixes part of <https://github.com/ShiftLeftSecurity/product/issues/5091>.


fabsx00:
So, was there a time when we used "GOLANG"? If not, then you can also just change that constant.

Ferada:
> So, was there a time when we used "GOLANG"? If not, then you can also just change that constant.

There wasn't in `go2cpg`. Okay, if it's not used I can change it.

PR #863, opened 2020-07-24
Modify how contained_by works

bbrehm:
Central discussion thread is https://github.com/ShiftLeftSecurity/codescience/issues/4038


codepropertygraph/src/main/scala/io/shiftleft/passes/DiffGraphProtoSerializer.scala
```diff
@@ -150,12 +148,20 @@ class DiffGraphProtoSerializer {
       .setPropertyName(EdgePropertyName.valueOf(propertyKey))
       .build
 
-  private def nodeProperty(key: String, value: Any) = {
-    CpgStruct.Node.Property
-      .newBuilder()
-      .setName(NodePropertyName.valueOf(key))
-      .setValue(protoValue(value))
-      .build()
+  //fixme
```
ml86, line 151: ???

codepropertygraph/src/main/scala/io/shiftleft/passes/DiffGraphProtoSerializer.scala
```diff
@@ -136,11 +124,21 @@ class DiffGraphProtoSerializer {
...
       .build
 
-  private def removeNodePropertyProto(nodeId: Long, propertyKey: String) =
-    DiffGraphProto.RemoveNodeProperty.newBuilder
-      .setKey(nodeId)
-      .setName(NodePropertyName.valueOf(propertyKey))
-      .build
+
+  private def removeNodePropertyProto(nodeId: Long, propertyKey: String) = {
+    if (propertyKey(0).isLower) {
```
ml86, line 129: This disguishment looks fishy. Having a property start with lower case to encode it as CONTAINED_REF is a way to implicit encoding.

codepropertygraph/src/main/scala/io/shiftleft/passes/DiffGraphProtoSerializer.scala
```diff
@@ -150,12 +148,20 @@ class DiffGraphProtoSerializer {
...
 
-  private def nodeProperty(key: String, value: Any) = {
-    CpgStruct.Node.Property
-      .newBuilder()
-      .setName(NodePropertyName.valueOf(key))
-      .setValue(protoValue(value))
-      .build()
+  //fixme
+  private def nodeProperty(key: String, value: Any, appliedDiffGraph: AppliedDiffGraph) = {
+    if (key(0).isLower) {
```
ml86, line 153: This disguishment looks fishy. Having a key start with lower case to encode it as CONTAINED_REF is a way to implicit encoding.

codepropertygraph/src/main/scala/io/shiftleft/passes/DiffGraphProtoSerializer.scala
bbrehm: :joy: 

It is fixed now, this is a stale memo-at-self from developing this (rather voluminous and coupled) PR. I will remove the comment in the next improvement round of this PR.

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/passes/languagespecific/fuzzyc/MethodStubCreator.scala
```diff
@@ -13,44 +13,26 @@ case class NameAndSignature(name: String, signature: String)
...
-        createMethodStub(name, name, signature, parameterCount, dstGraph)
-      case _ =>
     }
-    Iterator(dstGraph.build())
+    val builder = DiffGraph.newBuilder
+    for ((NameAndSignature(name, signature), nargs) <- methodToParameterCount) {
+      createMethodStub(name = name, fullName = name, signature = signature, parameterCount = nargs, dstGraph = builder)
+    }
+    Iterator(builder.build)
   }
```
ml86, line 37: This change seems unrelated.
The current version does not seem perfect. For example it iterates over a map which does not quarante and ordering but so does the new version.
So please explain why you need this change and why it is in this PR.

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/passes/languagespecific/fuzzyc/MethodStubCreator.scala
```diff
@@ -13,44 +13,26 @@ case class NameAndSignature(name: String, signature: String)
...
-      .sideEffect { method =>
-        methodFullNameToNode += method.fullName -> method
-      }
-      .exec()
-
-    cpg.call
-      .sideEffect { call =>
-        methodToParameterCount +=
-          NameAndSignature(call.name, call.signature) -> call._astOut.asScala.size
+  private val methodFullNameToNode = scala.collection.mutable.HashMap[String, nodes.MethodBase]()
```
ml86, line 20: Please stop using fully qualified class names. There is a reason why the `import` statement exists...
So please use `mutable.HashMap` like most of the other code.

codepropertygraph/src/test/scala/io/shiftleft/passes/CpgOverlayIntegrationTest.scala
```diff
@@ -42,15 +41,12 @@ class CpgOverlayIntegrationTest extends WordSpec with Matchers {
...
-            override def label = NodeTypes.UNKNOWN
-            override def properties = Map.empty
-          }
+          nodes.NewUnknown(code = null)
         ))
       cpg.graph.nodeCount shouldBe 2
-      val additionalNode = cpg.graph.V.hasNot(NodeKeysOdb.CODE).head
+      val additionalNode = cpg.graph.V.label("UNKNOWN").head
 
+      /* we don't need to support edges with properties any longer
```
ml86, line 49: That is not yet true. So please dont just comment this out.

codepropertygraph/src/main/scala/io/shiftleft/passes/DiffGraphProtoSerializer.scala
bbrehm: It _is_ super fishy!

The correct approach would be to check whether the property exists in the enum. However, afaiu proto doesn't have an api for that (unless we want to use exceptions for regular control flow, but that would be even uglier)

The even better thing would be to have "real" proto properties -- but that would require us to remodel everything within the same mindset of "cpg-style" properties (i.e. properties exist and are documented independently of the nodes that carry them -- as opposed to "class-style" properties, where it is conceptually nonsensical to even think about two classes with the same property, unless the property is declared in a common ancestor in the inheritance hierarchy). If I start remodeling the object hierarchy, then I'll vanish for weeks and rewrite too much code (and @fabsx00 would rightfully crucify me for exhibiting not-invented-by-myself syndrome), so I tried to stick to rewriting only code that is either both objectively awful and a bad offender in profiling, or needs to be rewritten to be compatible with the changes.

Do you have other better ideas? Did I just fail to rtfm java protobuf?

fabsx00:
Took the liberty to also assign @mpollmeier for review as he knows a lot about this code, and in particular the DiffGraphProtoSerializer, I believe.

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/passes/languagespecific/fuzzyc/MethodStubCreator.scala
bbrehm: Fair enough, I'll try to move the change to a separate PR (some intermediate variant of this PR during development failed tests here).

codepropertygraph/src/test/scala/io/shiftleft/passes/CpgOverlayIntegrationTest.scala
bbrehm: How?

This PR effectively deprecates CONTAINED_BY edges, and the only remaining use is PROPAGATE edges with the ALIAS flag in the open source dataflow tracker. However, the alias flag is always set to false, so no problem removing support for that.

bbrehm: Ok, I made it true by deprecating every single edge property that is ever used. Now good?

bbrehm:
Markus asked me to explain the changes from a high level.

We now store contained nodes as proper properties. In order to make this possible, this reworks a lot of the generated code (overflowdb-codegen) as well as the diffgraph applier. Downstream and upstream of that are minor api changes.

First of all, NewNode is now a mutable class, instead of a `case class`. I see no reason why this ever was a case class, since NewNode always behaves by-identity, and never by-equality. That causes some minor changes in codescience.

When applying a new node, there was never a reason to go detours via string-based hashmaps and add one property after the other. Instead, we just need an API function to initialize a StoredNode's instance fields from a matching NewNode.

Next, the old storeRecursively step was pretty expensive since it needed to traverse referenced nodes, check them against a dictionary (custom hashmap because we used overrode `equals` for no good reason), etc. This is unneeded complexity. When we initialize a node from NewNode, then we need to translate referenced NewNodes into StoredNodes. Because of the possibility of cycles, we use the usual pattern of "allocate StoredNode node first, then initialize them" -- i.e. if we spot a NewNode that is reachable from a NewNode that we try to store, then we need to store that one as well, with deferred initialization.

So, storeRecursively is now trivial, and the recursive structure is unwound during application of the diffgraph. The same holds for edges, so passes / frontends don't need to care about ordering anymore (it is now ok to add an edge to a newNode before adding the NewNode -- in fact, the edge makes the addNode redundant). 

The internal API for diffGraph serialization and deserialization uses the same protobuf as node storage. So the proto needs to somewhat deal with the new node properties as well. For this, I added a new property for "containedRef" -- and here, the value has an additional string that describes the "localName". Then we use that to refer to the field. 

Furthermore, overflowdb deserialization used to hand List-valued properties one-after-the-other. This was capital-B BAD, cf https://github.com/ShiftLeftSecurity/codescience/issues/4038

codepropertygraph/src/main/scala/io/shiftleft/passes/DiffGraph.scala
```diff
@@ -380,21 +407,28 @@ object DiffGraph {
...
+    private def addNodeNoInit(node: NewNode): Unit = {
+      if (overlayNodeToOdbNode.get(node) == null) {
+        val newNode = keyPool match {
+          case Some(pool) => graph.addNode(pool.next, node.label).asInstanceOf[StoredNode]
+          case None       => graph.addNode(node.label).asInstanceOf[StoredNode]
+        }
+        inverseBuilder.onNewNode(newNode.asInstanceOf[StoredNode])
+        overlayNodeToOdbNode.put(node, newNode)
+      } else {
+        throw new RuntimeException("something went wrong in diffgraph application")
```
mpollmeier, line 430: please add a bit more context so that if someone on call is woken up at 3am and reads this has at least a slight clue, or otherwise add your 24/7 contact number

codepropertygraph/src/main/scala/io/shiftleft/passes/DiffGraphProtoSerializer.scala
```diff
@@ -59,11 +61,13 @@ class DiffGraphProtoSerializer {
...
     import DiffGraph.Change._
     val builder = DiffGraphProto.newBuilder
+
     def newEntry = DiffGraphProto.Entry.newBuilder
+
     diffGraph.iterator
       .map {
         case SetNodeProperty(node, key, value) =>
-          newEntry.setNodeProperty(addNodeProperty(node.id2, key, value))
+          newEntry.setNodeProperty(addNodeProperty(node.id2, key, value, null))
```
mpollmeier, line 70: If the appliedDiffGraph is only sometimes passed, please pass it as an `Option` rather than handling `null` around. 
This makes the code more explicit ("it may not be set") and avoids spurious NPE that may be hard to debug.

codepropertygraph/src/main/scala/io/shiftleft/passes/DiffGraphProtoSerializer.scala
mpollmeier: > The correct approach would be to check whether the property exists in the enum. However, afaiu proto doesn't have an api for that

How about something like `io.shiftleft.proto.cpg.Cpg.NodePropertyName.values.toSet` (and then hold onto it somewhere as long as you need it)?

codepropertygraph/src/main/scala/io/shiftleft/passes/DiffGraphProtoSerializer.scala
```diff
@@ -150,12 +164,22 @@ class DiffGraphProtoSerializer {
...
 
-  private def nodeProperty(key: String, value: Any) = {
-    CpgStruct.Node.Property
-      .newBuilder()
-      .setName(NodePropertyName.valueOf(key))
-      .setValue(protoValue(value))
-      .build()
+  private def nodeProperty(key: String, value: Any, appliedDiffGraph: AppliedDiffGraph) = {
+    //this way of distinguishing between "proper" and "localName"-style properties is super fishy. Apologies!
+    if (key(0).isLower) {
```
mpollmeier, line 169: same as above. too much :fish:

codepropertygraph/src/main/scala/io/shiftleft/passes/DiffGraphProtoSerializer.scala
```diff
@@ -182,6 +209,26 @@ class DiffGraphProtoSerializer {
...
+    value match {
+      case iterable: Iterable[_] =>
+        iterable.foreach { node =>
+          node match {
+            case storedNode: StoredNode =>
+              builder.addRefs(storedNode.id2)
+            case newNode: NewNode =>
+              builder.addRefs(appliedDiffGraph.nodeToGraphId(newNode))
+          }
+        }
```
mpollmeier, line 223:
```suggestion
        iterable.foreach {
          case storedNode: StoredNode =>
            builder.addRefs(storedNode.id2)
          case newNode: NewNode =>
            builder.addRefs(appliedDiffGraph.nodeToGraphId(newNode))
        }
```

codepropertygraph/src/test/scala/io/shiftleft/passes/CpgOverlayIntegrationTest.scala
bbrehm: Resolving: We decided to do both, i.e. deprecate all existing edge properties and maintain tests and code for edge properties, on the assumption that we will need it again.

bbrehm:
This, as well as both upstream PRs are ready for re-review. Please disregard the `build.sbt` changes for that -- they are not final and will not be merged like here, they are just for testing (since the branch doesn't have version numbers)

bbrehm:
I'll refrain from merging this until the current sonatype troubles are resolved.

mpollmeier:
while i do not have an explanation for the travis/sonatype issues, we've so far only seen them with fuzzyc

PR #865, opened 2020-07-25
step: modifier.modifierType

mpollmeier:
cpg.method.modifier.modifierType

In future we should autogenerate those again, IMO.


mpollmeier:
cc @ltcmelo

fabsx00:
+:1: on autogenerate: if the `name` in `cpg.method.map(._name)` is auto-generated, then so should `cpg.method.name` be. Also see https://github.com/ShiftLeftSecurity/codescience/issues/4018 . This issue has also been pointed out by customers, and keeping an up-to-date manually maintained version of accessors has proven to not work over time.

mpollmeier:
https://github.com/ShiftLeftSecurity/overflowdb-codegen/pull/26

PR #869, opened 2020-07-27
rework fuzzyC-methodStub creator

bbrehm:


fabsx00:
Is this ready for review and can't you explain what you're doing and why?

bbrehm:
Sorry for the missing explanation! It is ready for review.

I pulled this out of https://github.com/ShiftLeftSecurity/codepropertygraph/pull/863

This reworks the fuzzyC methodStub creator. The old code is rather grown over time (per git history) by many people, and was therefore quite silly (when viewed on its own merits, i.e. without its history). During debugging of https://github.com/ShiftLeftSecurity/codepropertygraph/pull/863 I reworked this anyway, so better to merge the fruits of that labor than let it go to waste. The new code is much simpler and should be much faster.

fabsx00:
So, why are we not using `ParallelCpgPass` anymore? I benchmarked this pass with and without `ParallelCpgPass` last time I touched it and it was considerably faster when using `ParallelCpgPass`.

bbrehm:
I am 95% sure that this is a perf improvement, in addition to being a readability improvement. 

I'll post benchmarks later, and we'll keep this PR open until your doubts are assuaged.

fabsx00:
I'm getting daily reminders about this one, so, let's close it for now and reopen when the time presents itself.

PR #872, opened 2020-07-28
Add basic auth logic for CPGQLServer

ursachec:
This change introduces a security mechanism for the
CPGQLServer, namely basic authentication. CPGQL queries that are
executed by the server have access to a large range of system resources,
having a way to restrict their execution seems very useful in a lot of
situations.

The implementation is bare-bone, but does the job. A few caveats:
1. The `/connect` endpoint has not been secured, mostly because the
`cask` library does not yet provide a simple way of adding decorators to
WebSocket endpoints.
2. 401 responses do not return a body, simply because the `requests`
library returns an empty buffer for them, even though the HTTP standard
allows it.
3. The basic auth logic has been implemented by hand because the `cask`
library does not provide a decorator out of the box, and doing it using
the underlying `undertow` library would probably require too much code:
ihttps://github.com/undertow-io/undertow/blob/master/examples/src/main/java/io/undertow/examples/security/basic/BasicAuthServer.java

### DIFF
--- a/console/src/main/scala/io/shiftleft/console/BridgeBase.scala
+++ b/console/src/main/scala/io/shiftleft/console/BridgeBase.scala
@@ -127,7 +127,11 @@ trait BridgeBase {
     Runtime.getRuntime.addShutdownHook(new Thread(() => {
       ammonite.shutdown()
     }))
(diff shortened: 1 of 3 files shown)


fabsx00:
Oh, please update the documentation accordingly.

PR #876, opened 2020-08-01
Adapt to changes in fuzzyc

fabsx00:
Re: https://github.com/ShiftLeftSecurity/fuzzyc2cpg/pull/222


semanticcpg/src/main/scala/io/shiftleft/semanticcpg/passes/linking/filecompat/FileNameCompat.scala
```diff
@@ -18,7 +18,8 @@ class FileNameCompat(cpg: Cpg) extends CpgPass(cpg) {
     val dstGraph = DiffGraph.newBuilder
 
     def updateDefaultFileName(node: nodes.StoredNode with nodes.HasFilename): Unit = {
-      if (node.filename == null) {
+      // When creating nodes via NewNode classes, filename is "", not null.
+      if (node.filename == null || node.filename == "") {
```
fabsx00, line 22: This is the one change that may be problematic, so, if we see trouble down the line, this may need to be reverted.

PR #886, opened 2020-08-05
Move over `KeyPoolCreator` from fuzzyc2cpg

fabsx00:
All KeyPool code is in `codepropertygraph`, except for a handy utility method called `KeyPools.obtain`. Copied this one over from fuzzyc because I need it in other frontends, also added missing test.


ursachec:
:dancers: @tuxology

PR #888, opened 2020-08-07
Round of cleanups for ReachingDefPass

fabsx00:
This is a first round of cleanups for the ReachingDefPass to make the algorithm more visible in preparation for its improvement. It also turned out that we computed the `gen` map a few times although it is sufficient to calculate it only once. This has been fixed, so, the change should also improve performance. There is still a lot of room for improvement here, but I thought I'd make @mpollmeier aware of these changes given that he's currently working on a larger refactor.

### DIFF
--- a/dataflowengineoss/src/main/scala/io/shiftleft/dataflowengineoss/passes/reachingdef/ReachingDefPass.scala
+++ b/dataflowengineoss/src/main/scala/io/shiftleft/dataflowengineoss/passes/reachingdef/ReachingDefPass.scala
@@ -11,25 +11,44 @@ import io.shiftleft.semanticcpg.utils.MemberAccess
 import scala.collection.mutable
 import scala.jdk.CollectionConverters._
 
+/**
+  * A pass that calculates reaching definitions ("data dependencies").
+  * */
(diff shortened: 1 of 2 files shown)


dataflowengineoss/src/main/scala/io/shiftleft/dataflowengineoss/passes/reachingdef/ReachingDefPass.scala
```diff
@@ -231,36 +269,19 @@ object DataFlowFrameworkHelper {
   /** Returns a set of nodes that are killed by the passed nodes */
   def killsVertices(node: nodes.StoredNode): Set[nodes.StoredNode] = {
 
-    val localRefIt = node._refOut.asScala
+    val localRefIt = node._refOut().asScala
```
mpollmeier, line 272: personally i prefer to omit the noise of `()` everywhere, but ¯\_(ツ)_/¯

fabsx00: intellij wouldn't shut up about it :/

PR #891, opened 2020-08-10
Add more convenient accessor for arguments of calls

bbrehm:
Add more convenient accessor argumentOption for call->argument.  Make old accessor faster (don't use gremlin).

This is more convenient because many possible uses want to handle the None-case instead of having an exception thrown in their face. Ping @ml86 to make you aware of this new API (it can simplify a lot of code in the closed source dataflow tracker!).

### DIFF
--- a/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/nodemethods/CallMethods.scala
+++ b/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/nodemethods/CallMethods.scala
@@ -5,9 +5,10 @@ import io.shiftleft.semanticcpg.language._
 import scala.jdk.CollectionConverters._
 
 class CallMethods(val node: nodes.Call) extends AnyVal {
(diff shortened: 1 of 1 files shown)


semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/nodemethods/CallMethods.scala
```diff
@@ -2,7 +2,12 @@ package io.shiftleft.semanticcpg.language.nodemethods
 
 import io.shiftleft.codepropertygraph.generated.nodes
 import io.shiftleft.semanticcpg.language._
+import scala.jdk.CollectionConverters._
 
 class CallMethods(val node: nodes.Call) extends AnyVal {
-  def argument(index: Int): nodes.Expression = node.start.argument(index).head
+  def argumentOption(index: Int): Option[nodes.Expression] =
```
ml86, line 8: Lets go for `Iterable` instead of `Option` as return type. That way we might return vargs later.

PR #909, opened 2020-08-18
Assume that calls to unannotated methods redefine all args

fabsx00:
This is a pretty important change for the data flow engine. Consider the following code:

```
int foo (int x) {
   bar(x);
   sink(x);
}
```

and requests to determine flows from the parameter declaration `int x` to the argument `x` of the call to sink. Consider also that `bar` has NOT been annotated, that is, we know nothing about the way it propagates taint. If we now assume by default that calls to unannotated methods such as `bar(x)` do not lead to modifications of the argument (`x`), then the flow that comes back is

```
int x
sink(x)
```

If we assume that it does change its argument, we arrive at
```
int x
bar(x)
sink(x)
```
For exploratory code analysis, I argue that the latter is better because we end up with flows that show us all transformations that _may_ affect `x`. It's possible that flows with `bar(x)` on the path are also uninteresting for us, in which case we an filter for flows that do not contain `bar(x)`, where is with the previous behavior, we wouldn't know whether `bar(x)` is on the flow or not. 

Annotation, with this default behavior, leads to less flows being returned, but annotation is not a prerequisite for flows to be returned at all. The downside, of course, is that our flows may be much too detailed, but again, in exploratory mode, this makes sense and can then be compensated by filtering flows.

### DIFF
--- a/dataflowengineoss/src/main/scala/io/shiftleft/dataflowengineoss/passes/reachingdef/ReachingDefPass.scala
+++ b/dataflowengineoss/src/main/scala/io/shiftleft/dataflowengineoss/passes/reachingdef/ReachingDefPass.scala
@@ -56,7 +56,8 @@ class ReachingDefPass(cpg: Cpg) extends ParallelCpgPass[nodes.Method](cpg) {
                     .filter(_.isInstanceOf[nodes.CfgNode])
                     .map(_.asInstanceOf[nodes.CfgNode].code)
                     .getOrElse("")
-                  addEdge(in, use, edgeLabel)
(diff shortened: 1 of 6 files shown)


ltcmelo:
Is this [related to @bbrehm 's proposal](https://github.com/ShiftLeftSecurity/codescience/pull/4085) ? @bbrehm @fabsx00 @ml86

fabsx00:
No, this is the oss data flow tracker.

ltcmelo:
> No, this is the oss data flow tracker.

I'm mean conception-wise, not implementation-wise necessarily.

PR #911, opened 2020-08-19
Ditch obsolete log4j2 config

hubertp:
Any implementation should provide it's own config when running.
It's better than excluding that file every single time.


hubertp:
That will fix the problem in downstream projects

max-leuthaeuser (approved):

PR #912, opened 2020-08-19
Use mutable data structures in CfgDominator calculation.

ml86:
Only a small gain of a few seconds for very large CPGs but since it is
already changed we might as well take it.

### DIFF
--- a/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/passes/cfgdominator/CfgDominator.scala
+++ b/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/passes/cfgdominator/CfgDominator.scala
@@ -1,5 +1,7 @@
 package io.shiftleft.semanticcpg.passes.cfgdominator
 
+import scala.collection.mutable
+
 class CfgDominator[NodeType](adapter: CfgAdapter[NodeType]) {
 
   /**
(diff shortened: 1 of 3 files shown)


semanticcpg/src/main/scala/io/shiftleft/semanticcpg/passes/cfgdominator/CfgDominator.scala
```diff
@@ -87,10 +89,10 @@ class CfgDominator[NodeType](adapter: CfgAdapter[NodeType]) {
     finger1
   }
 
-  private def createPostOrderNumbering(cfgEntry: NodeType): Map[NodeType, Int] = {
+  private def createPostOrderNumbering(cfgEntry: NodeType): mutable.Map[NodeType, Int] = {
     var stack = (cfgEntry, adapter.successors(cfgEntry).iterator) :: Nil
-    var visited = Set.empty[NodeType]
-    var numbering = Map.empty[NodeType, Int]
+    var visited = mutable.Set.empty[NodeType]
```
bbrehm, line 94: If we're touching this code anyways, may as well `val visited = ...`

bbrehm (approved):
These changes are fine. If we find the opportunity, I'd like to move the DFO-Pass from codescience to codepropertygraph and merge it with this.

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/passes/cfgdominator/CfgDominator.scala
ml86: Done.

ml86:
Yes, reusing the DFO makes sense at some point.

PR #926, opened 2020-08-31
Fix method filename in locations.

Ferada:
Fixes <https://github.com/ShiftLeftSecurity/product/issues/5393#issuecomment-683800676>, the reported filenames are wrong, although it should now be possible to be more accurate since <https://github.com/ShiftLeftSecurity/codepropertygraph/pull/622> was merged.

Happy to know it can't be done this way, apart from that, it actually works for this case.

But, it doesn't seem to work e.g. in `sptests`, as the attribute doesn't get set? This shouldn't be the case since the `FileNameCompat` pass would set it if missing.

In any case, appreciate any input on this.

### DIFF
--- a/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/LocationCreator.scala
+++ b/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/LocationCreator.scala
@@ -16,10 +16,9 @@ object LocationCreator {
     try {
       location(node)
     } catch {
-      case exc @ (_: NoSuchElementException | _: ClassCastException) => {
+      case exc @ (_: NoSuchElementException | _: ClassCastException) =>
         logger.error(s"Cannot determine location for ${node.label} due to broken CPG", exc)
         emptyLocation(node.label, Some(node))
-      }
     }
   }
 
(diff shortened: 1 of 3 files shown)


semanticcpg/src/main/scala/io/shiftleft/semanticcpg/utils/ExpandTo.scala
```diff
@@ -62,17 +62,14 @@ object ExpandTo {
...
   def hasModifier(methodNode: nodes.StoredNode, modifierType: String): Boolean =
     methodNode._astOut.asScala
-      .exists(
-        astChild =>
-          astChild.isInstanceOf[nodes.Modifier] &&
-            astChild.asInstanceOf[nodes.Modifier].modifierType == modifierType)
+      .exists(astChild =>
+        astChild.isInstanceOf[nodes.Modifier] &&
+          astChild.asInstanceOf[nodes.Modifier].modifierType == modifierType
+      )
```
Ferada, line 68: `scalafmt`

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/LocationCreator.scala
```diff
@@ -112,11 +112,13 @@ object LocationCreator {
...
-            label: String,
-            lineNumber: Option[Integer],
-            method: nodes.Method): nodes.NewLocation = {
+  def apply(
+      node: nodes.CpgNode,
+      symbol: String,
+      label: String,
+      lineNumber: Option[Integer],
+      method: nodes.Method
+  ): nodes.NewLocation = {
```
Ferada, line 121: `scalafmt`

ml86 (approved):

mpollmeier (approved):

PR #938, opened 2020-09-04
fix multi-regex accessors

mpollmeier:


hubertp (approved):
That seems... expensive to do on every string value?
Maybe we should have a special version of `evalType` for that, or at least document it that it is possible to have a regex here

mpollmeier:
for all accessors we default to regexes, but they all have an `Exact` alternative as well: https://github.com/ShiftLeftSecurity/codepropertygraph/blob/2eff2c6a0b0692943835cfd11385682978e9765c/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/types/propertyaccessors/EvalTypeAccessors.scala#L33

PR #954, opened 2020-09-15
upgrade codegen: use auto-generated property accessors

mpollmeier:


fabsx00 (approved):

mpollmeier:
this needed a fuzzyc upgrade (which complexifies things due to cyclic dependencies). should be green shortly

PR #962, opened 2020-09-21
Cleanup (very) verbose logging in tests

hubertp:
Noticed that regular `sbt> test` produced a ton of logs that were hard
to follow and not even possible to configure. In order to improve the
situation:
1. Moved test fixtures code from `src/main` to `test`, since it is
really test code
2. This enabled me to make `fuzzyc2cpg` a test dependency rather than a
compile time dependency and forcing it on everyone
3. Excluded `logback` from all `fuzzyc2cpg` deps since for some unknown
reason it is the logging implementation of choice there. All other
projects only make log4j2 as a runtime dep, so someone should revisit
that
4. Added log4j2 config that is **only** used in tests (therefore not
being passed downstream)
5. Needed to fork tests in order to be able to assign the external
config, which brought its own problems when dealing with subprojects

As a result `sbt> test` produces much, much less noise and one is able
to control it via the `resources/log4j2-test.xml` file, if necessary.


mpollmeier (approved):
test output is indeed much nicer now. 
i double (spot-) checked that tests still work when run from intellij (since you changed the Test/fork setting)
:+1:

fabsx00:
This breaks https://github.com/ShiftLeftSecurity/joern-sample-extension, unfortunately. That was the reason for the test code being in `main`: if it is not, then it is not included in the jar. Can we generate a test jar maybe and include that?

fabsx00:
@hubertp actually, it also breaks `joern-latest-deps`. `ConsoleTests.scala` there relies on `io.shiftleft.console.testing.TestCpgGenerator`. I think we'll need to revert this.

hubertp:
I'm looking into it. Publishing test artifacts is not a problem and that's how it should be done in the first place.

fabsx00:
Ok, great.

PR #970, opened 2020-09-25
refactor tracking point methods

bbrehm:
In order to make simple alias handling reasonable, this refactors the various tracking point -> trackedBase and accessPath methods.

The underlying theme is that trackedBase and accessPath are only needed as a pair, and never in isolation (at least conceptually).

In order to simplify the changes in codescience, we also introduce the `FakeTrackingPoint` interface, which allows us to get rid of the `${foo}Ext` variants. 

Furthermore, this gets rid of the theoretical O(N^2) part of the computation. That was benign, though.

I decided to keep the function recursive instead of writing a `while`-loop for better readability + better stacktraces, and in order to optimize for the simple case (recursion ends after the first step). Otherwise, we'd need annoying `var accumulator: ArrayBuffer[AccessElement] = null` with corresponding checks, in order to avoid the allocation of the (typically empty) buffer.

Before merging, this also needs approval on the codescience side.


dataflowengineoss/src/main/scala/io/shiftleft/dataflowengineoss/passes/reachingdef/ReachingDefProblem.scala
```diff
@@ -125,29 +120,27 @@ class ReachingDefTransferFunction(method: nodes.Method) extends TransferFunction
...
-          val otherCallsWithSameBase = baseToCalls
-            .get(base)
-            .toList
-            .flatten[nodes.Call]
-            .filter(_.id != node.id)
-          otherCallsWithSameBase
-            .filter { n =>
-              val (matchResult, _) = TrackingPointToAccessPath(n).matchAndDiff(accessPath.elements)
-              matchResult == MatchResult.EXACT_MATCH
+          val (base, accessPath) = TrackingPointMethods.toTrackedBaseAndAccessPath(call)
```
fabsx00, line 134: If we only ever want to work with the tracker base and accessPath at once, I'd suggest we wrap them in a class and give that class a descriptive name. I would actually imagine that we can get away with calling the whole thing `AccessPath`.

fabsx00: Let's then also introduce a corresponding `apply` method, simply because `TrackingPointMethods.to...` is pretty long.

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/accesspath/TrackingPointToAccessPath.scala
```diff
@@ -9,101 +9,180 @@ import org.slf4j.LoggerFactory
 
 import scala.jdk.CollectionConverters._
 
-object TrackingPointToAccessPath {
-
-  def apply(node: nodes.TrackingPoint, exclusions: List[Elements] = List()): AccessPath = {
-    new AccessPath(TrackingPointToElements(node), exclusions)
-  }
-
+trait FakeTrackingPoint {
```
fabsx00, line 12: So, can you explain again why we need this fake tracking point? The reason I ask is that anyone reading this will ask themselves in which way this tracking point is fake. The reason I didn't originally port over these fake tracking points is because it seemed conceptually unclean to me.

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/accesspath/TrackingPointToAccessPath.scala
```diff
@@ -9,101 +9,180 @@ import org.slf4j.LoggerFactory
...
+                     .collect {
+                       case lit: nodes.Literal => ConstantAccess(lit.code)
+                     }
+                     .getOrElse(VariableAccess) :: tail)
+              }
+              .getOrElse {
+                logger.warn(s"Missing argument on call ${memberAccess}.")
+                (TrackedUnknown, Nil)
+              }
+          case Operators.indirection =>
```
fabsx00, line 111: Looks like there's a lot of duplication here and the function is very long. Any way we can reduce the amount of code here?

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/accesspath/TrackingPointToAccessPath.scala
bbrehm: They _are_ conceptually unclean -- the underlying architecture wants all subclasses of `node` to be generated and known to odb, i.e. these should really be `sealed`.

However, we have several APIs that expect `node` that turned out to sometimes require extra data, and therefore we have a bunch of fake nodes all over the place (most prevalent in the closed source data flow tracker) -- that is, classes that claim to implement out node interfaces, but in reality only implement a tiny subset, and will blow up if they get serialized (e.g. because they are part of a Flow).

For this reason, codescience currently needs its own implementation (API entry-point) for all these methods, in order to `match` over the fake node classes. An effect of that is that it is hard to share code between the open and closed tracker (in the open source world, you use these methods for accessing base+path, in the closed source world you use these completely different methods).

That is bonkers, so it's better to define one extra interface for fake tracking points and use that. Yes, it's unclean, but this way we at least contain the uncleanliness.

I'll add source code comments to that effect.

dataflowengineoss/src/main/scala/io/shiftleft/dataflowengineoss/passes/reachingdef/ReachingDefProblem.scala
bbrehm: I'm not sure. The object-apply pattern is super ugly for what is effectively a free-standing function.

What I think would be preferred is an implicit, such that we can e.g. `val (base, path) = someNode.trackedBaseAndPath` or `val (base, elems) = someNode.trackedBaseAndPathElems`, where `trackedBaseAndPathElems(node) = {val (base, path) = trackedBaseAndPath(node); (base, path.elements)}`.

The `trackedBaseAndPathElems` would be for the closed tracker -- large parts of it can only deal with `Elements`, not full paths. That's OK in its context, but typing it again and again gets old fast.

The big question about that is whether to expose the implicit to users, i.e. whether to put it into some hidden obscure module (beware of the leopard!), or put it into one of our predef language extension packages.

ml86 (approved):

PR #976, opened 2020-09-27
Performance fixes for `expandIn`

fabsx00:
Running DDG extraction on different samples uncovered a performance problem: `expandIn` was recalculating the same results many times where installing a cache was simple. This PR brings in a cache.


dataflowengineoss/src/main/scala/io/shiftleft/dataflowengineoss/queryengine/Engine.scala
```diff
@@ -125,12 +124,45 @@ class Engine(context: EngineContext) {
     forCalls ++ forArgs
   }
 
+  private def paramToArgs(param: nodes.MethodParameterIn): List[nodes.Expression] =
```
mpollmeier, line 127: for more consistency: should all these defs return only one of `List|Traversal`, rather than a wild mix?

mpollmeier (approved):
i know you didn't really change what i commented on, but still :)

PR #978, opened 2020-09-27
Bump Ammonite version

hubertp:
Downstream, internally, we are already using `2.0.4`, so the upgrade
should be relatively painless.
Postponing the upgrade to `2.2.0` since it would require bumping Scala
version (`2.13.2` or `2.13.3`) as well.

Test changes are cosmetic due to API/implementation changes of ammonite.


mpollmeier (approved):

fabsx00:
Please also check manually whether completion still works, e.g., `works<TAB>` should complete to `workspace`.

mpollmeier:
tested, works

PR #984, opened 2020-10-05
resolve cyclic cpg<->fuzzyc dependency by using fuzzyc binary

mpollmeier:
* instead of having both in the same path we now use fuzzyc as a separate subproject with it's own classpath
* all dependent subprojects ensure that fuzzyc is staged before they run their tests
* works from sbt and intellij
* once this is approved/merged i'll apply the same mechanism to the codescience build, which has the same issue


fabsx00 (approved):
Good for now. I'm still in favor of just merging `fuzzyc` into `codepropertygraph` entirely over time, but this is as good as it gets right now.

PR #988, opened 2020-10-08
Add POST_EXECUTION_CALL node for library modeling.

ml86:
A POST_EXECUTION_CALL node indicates the existence of a call executed
on a return value or out parameter of a method after this method has
been executed. This is used to model framework code calling functors
returned from user code. The outgoing REF edge indicates on which
returned entitity the call will happen.

### DIFF
--- a/codepropertygraph/src/main/resources/schemas/enhancements-internal.json
+++ b/codepropertygraph/src/main/resources/schemas/enhancements-internal.json
@@ -622,7 +622,10 @@
         {
             "name": "IMPLICIT_CALL", "keys" : ["DEPTH_FIRST_ORDER", "INTERNAL_FLAGS"]
         },
-        {
+      {
+        "name": "POST_EXECUTION_CALL", "keys" : ["DEPTH_FIRST_ORDER", "INTERNAL_FLAGS"]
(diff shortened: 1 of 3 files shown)


bbrehm:
Is this really necessary? Can't we use ImplicitCall nodes for that?

ImplicitCall is not a "real" node anyway, and looked up in SemanticsDatabase. We could simply augment SemanticsDatabase to handle this (just a preCall | postCall boolean flag). Or we could add a boolean flag to the node itself.

The reason I'm aksing is that everything that goes into the graph becomes slow, almost impossible to deprecate, and inflexible. 

Furthermore it adds a lot of mental overhead: Ideally everyone working with CPG should be able to hold the entirety of the spec + meaning  in their mind at all times -- including all the invariants that cannot be expressed as a graph spec. That is obviously not the case, but the extra complexity appears unneeded here.

bbrehm (approved):
Please ping me again when you're ready to merge -- I'd like to see the downstream changes before bringing this in.

ml86:
I am merging this now despite being red in order to break cyclic dependency with fuzzyc2cpg.

PR #991, opened 2020-10-09
Add POST_EXECUTION_CALL node for library modeling.

ml86:
A POST_EXECUTION_CALL node indicates the existence of a call executed
on a return value or out parameter of a method after this method has
been executed. This is used to model framework code calling functors
returned from user code. The outgoing REF edge indicates on which
returned entitity the call will happen.


mpollmeier:
https://github.com/ShiftLeftSecurity/codepropertygraph/pull/992 fixes this

bbrehm (approved):

PR #992, opened 2020-10-12
import fuzzyc into the cpg build

mpollmeier:
Resolves cyclic dependency between the two repos, and allows to modify the
schema again (unlike with the old version that used a staged fuzzyc
version)

Notes
* travis configuration will need to be verified on master, we may need additional fixup PRs
* I needed to extract semanticcpg-tests into a separate project to resolve a supposedly cyclic dependency (or what sbt claimed to be cyclic)

## new version number?
fuzzyc will now be released under the same version as cpg, which is currently 0.12.x. This will lead to conflicts in future, because fuzzyc is currently versioned as 1.1.x. 
To avoid those future problems, we should probably change the cpg version to 1.2.x now - thoughts?


ml86:
We definitely need to increase the cpg version to at least 1.2 yes

ml86 (approved):

fabsx00 (approved):

mpollmeier:
git submodules don't like me :(

PR #994, opened 2020-10-12
RFC: MustAlias edge and pass

bbrehm:
This introduces a new MUST_ALIAS edge, and a simple pass, to handle soot-generated aliases.

Before merging, consider the relevant changes in codescience and the resultant changes in findings.

But you can comment on the general approach already, and whether this justifies the additional edge.

A further change is to treat `<operator>.cast` like member-access. I think that would be a very sane change -- in JVM languages, we have a lot of cases like
```
$r1 = (java.util.Map) $r0 // $r0 has static type java.util.HashMap
```
Now we already have two mechanisms for dealing with aliases: First, all member access operators are treated as aliases-with-path by the entire machinery, deep down; second, we allow policies to declare the `-alias` flag, which triggers a slightly sketchy alias mechanism. We currently use that slightly sketchy mechanism for `<operator>.cast`. However, this mechanism does not interact well with the new must_alias pass.

I think we should always have treated `<operator>.cast` like a member access, for the sake of aliasing. That does not introduce any possible incorrect flows.

The incorrect flows caused by this new must_alias pass are all of the same kind: If a memory write between creation and use of the alias breaks it, then we compute wrong flows. I think this is preferable to the current state (we _always_ assume that a memory write in the meantime destroys the alias).

A more advanced pass would figure out whether intermediate (in CFG) memory writes can break the alias. We already do that in the llvm inliner pass, using the built-in alias analysis and callgraph-based function attributor (esp. readonly, argmemonly).

A similar approach could be attempted for CPG, but for now I'd like to see comments on the AST-based variant.

A more brutal approach would also alias SSA values (single definition, many uses), but I want to start small.


ml86:
As discussed via Zoom: We cannot merge it like this because this would eliminate the cast operations from the data flow results which is important to have in languages like C since problematic casts are a hole group of security problems in this language.

bbrehm:
@ml86 Can you take a look again?

The insight behind the latest change is that we don't need to know user preferences about `<operator>.cast` in the general code -- we can simply look at presence of `mustAlias` edges.

re naming: What about `SAFE_SSA`? The edge after all signifies that something is single-static-assignment (which allows us to elide tracking the LOCAL), and that it is permissible to relocate between DEF and USE.

Furthermore, I think we should move this from enhancements.json into base.json and allow frontends to use that: LLVM has its own version of this computation, and could set these edges (llvm2cpg actually builds an entire forest, not just chains like the simple pass for JVM, and uses actual pointer analysis to determine the SAFE part).

Currently llvm2cpg just recursively inlines the DEF into all USES -- which means that we have potentially exponential blow-up of the size of the CPG (DEF0 is inlined into USE00 and USE01, which are themselves define DEF1 which has USE10 and USE11, et cetera. In other words, `int x0 = foo(); int def0 = x0+x0; int def1 = def0 + def0; int def2 = def1 + def1; ...` leads to exponential number of copies of `def0`).

bbrehm:
@ml86 @fabsx00 can you take a look again? This work is not high priority, but it has been stuck in review for ~ 3 weeks now.

ml86:
@bbrehm I am still not convinced that sometimes treating the cast operator as part of the access path is a good idea.

bbrehm:
@ml86 it is the correct thing to do in java, where casts are pointer-reinterpret, and added by the compiler due to the way generics / type parameters are implemented.

It would have been better to model NOP-casts (i.e. reinterpret / bitcast of pointers) with a different operator in the first place, but that's where we are.

ml86 (approved):

fabsx00:
Closing, because I've been getting daily reminders about it for the last 6 months.

PR #995, opened 2020-10-12
Make `HASH` field optional

ursachec:


fabsx00 (approved):
This will work, but I'll also need to adapt java2cpg (which I will once this is merged).

PR #1016, opened 2020-11-06
Add `identifier` Node-Type-Step to `literal`

ursachec:


fabsx00:
Hm, I don't quite understand. Can you give an example?

ursachec:
@fabsx00 I was writing a custom pass and the situation arose in which it would have been useful to have access to paths through the CFG from LITERAL to IDENTIFIER, but @ml86 just explained to me that a `cfgPrev` does the job more clearly and that the step doe not make much sense. So closing...

PR #1017, opened 2020-11-09
Allow extensions from `io.shiftleft` and `io.joern` packages

fabsx00:
We were previously looking for extensions only in the `io.shiftleft` package.


fabsx00:
This casts the net to wide, leading to warnings when starting ocular/joern. I'll try to find a way to limit this to `io.shiftlefft` and `io.joern`.

prabhu (approved):

PR #1020, opened 2020-11-12
remove ambiguous implicits

mpollmeier:
prior to this we had two conversion paths from `nodes.TrackingPoint`
to `AstNode[nodes.AstNode]`, which would lead to ambiguous implicits
that don't resolve, e.g. for `cpg.call.where(_.argument.order(1).isIdentifier)`

re https://github.com/ShiftLeftSecurity/joern/issues/382


mpollmeier:
cc @jaiverma

fabsx00 (approved):

PR #1034, opened 2020-12-03
Add GitHub Actions setup

ursachec:
The *test jobs* (_test-codepropertygraph_, _test-fuzzypp-linux_, _test-fuzzypp-windows_) work with the exception of three test cases in _test-fuzzypp-windows_ that fail because of temporary file path problems (currently a mistery as to why they behave differently on Windows, so they're simply skipped for the moment).


fabsx00 (approved):

.github/workflows/pr.yml
```diff
@@ -0,0 +1,72 @@
+name: CI pipeline
+
+on:
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  test-codepropertygraph:
+    runs-on: ubuntu-latest
```
mpollmeier, line 9:
```suggestion
    runs-on: ubuntu-18.04
```
just a suggestion though - personally i'm in favor of controlling the environment myself, i.e. avoid surprising version updates by other parties. same goes for other versions in this file of course

.github/workflows/push.yml
```diff
@@ -0,0 +1,141 @@
...
+        run: "./fuzzypp/bin/Release/fuzzyppcli-tests.exe"
+      - name: Zip artifact
+        run: 7z a -r -tzip ./fuzzyppcli-win.zip ./fuzzypp/bin/Release/fuzzyppcli.exe
+      - name: Upload zip artifact
+        uses: actions/upload-artifact@v2
+        with:
+          name: windows-zip
+          path: ./fuzzyppcli-win.zip
+
+  release:
```
mpollmeier, line 74: i ported https://github.com/ShiftLeftSecurity/sbt-ci-release-early to use github-actions, which actually makes things easier. The readme is up to date and there's a few working examples already, e.g. https://github.com/ShiftLeftSecurity/overflowdb/blob/master/.github/workflows/release.yml

I'll try to suggest a workflow config, but since it's for release we'll need to merge to test it i guess.

.github/workflows/push.yml
```diff
@@ -0,0 +1,141 @@
...
+      - name: Tag release
+        run: sbt ciReleaseTagNextVersion
+      - name: Decrypt private-key.pem
+        run: openssl aes-256-cbc -K ${{ env.KEY_663dec309dc8 }} -iv ${{ env.IV_663dec309dc8 }} -in private-key.pem.enc -out private-key.pem -d
+      - name: Import GPG key
+        run: gpg --import --batch private-key.pem
+      - name: Publish `codepropertygraph` release to Sonatype
+        run: sbt ciReleaseSonatype
+      - name: Publish `codepropertygraph` release to ShiftLeft artifactory
+        run: sbt 'set publishTo := Some("releases" at "https://shiftleft.jfrog.io/shiftleft/libs-release-local")' "set credentials += Credentials(\"Artifactory Realm\", \"shiftleft.jfrog.io\", \"${{ secrets.JFROG_USER }}\", \"${{ secrets.JFROG_PASS }}\")" ciRelease
```
mpollmeier, line 105:
```suggestion
      - run: sbt test ciReleaseTagNextVersion ciReleaseSonatype
        env:
          PGP_PASSPHRASE: ${{ secrets.PGP_PASSPHRASE }}
          PGP_SECRET: ${{ secrets.PGP_SECRET }}
          SONATYPE_PASSWORD: ${{ secrets.SONATYPE_PASSWORD }}
          SONATYPE_USERNAME: ${{ secrets.SONATYPE_USERNAME }}
```

Those secrets are configured for the entire organization, so :crossed_fingers: this may just work. 
The following 'create_release' step is probably not needed, but you may know better.

mpollmeier (approved):
try with my suggestions, it should simplify it and actually make the release work. 
oh, and you'll need to upgrade the release plugin in plugins.sbt, I'll push this straight to your branch now

mpollmeier:
Also, i just pushed the removal of .travis.yml, otherwise we'll got two competing release builds which may get us into trouble

.github/workflows/pr.yml
ursachec: Yes, good catch!

PR #1078, opened 2021-01-06
Default set of passes for unknown languages

itsacoderepo:


itsacoderepo:
Closing because there is an internal hierarchy of the passes. The PR don't make much sense

PR #1090, opened 2021-01-12
Add documentation members to Query

ursachec:
Example usage in the `query-database`:

```
object CopyLoops extends QueryBundle {

  @q
  def isCopyLoop(): Query = Query(
    name = "copy-loop",
    author = Crew.fabs,
    title = "Copy loop detected",
    description =
      """
        |For (buf, indices) pairs, determine those inside control structures (for, while, if ...)
        |where any of the calls made outside of the body (block) are Inc operations. Determine
        |the first argument of that Inc operation and check if they are used as indices for
        |the write operation into the buffer.
        |""".stripMargin,
    score = 2,
    startLine = sourcecode.Line(),
    traversal = { cpg =>
      cpg.assignment.target.isArrayAccess
        .map { access =>
          (access.array, access.subscripts.code.toSet)
        }
        .filter {
          case (buf, subscripts) =>
            val incIdentifiers = buf.inAst.isControlStructure.astChildren
              .filterNot(_.isBlock)
              .assignments
              .target
              .code
              .toSet
            (incIdentifiers & subscripts).nonEmpty
        }
        .map(_._1)
    },
    endLine = sourcecode.Line(),
    fileName = sourcecode.FileName()
  )

}
```

### DIFF
--- a/console/src/main/scala/io/shiftleft/console/QueryDatabase.scala
+++ b/console/src/main/scala/io/shiftleft/console/QueryDatabase.scala
@@ -20,7 +20,13 @@ case class Query(name: String,
                  title: String,
                  description: String,
                  score: Double,
+                 // intended to be filled by com.lihaoyi.sourcecode.Line
+                 docStartLine: Int = 0,
                  traversal: Cpg => Traversal[nodes.StoredNode],
+                 // intended to be filled by com.lihaoyi.sourcecode.Line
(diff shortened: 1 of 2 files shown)


tuxology (approved):

fabsx00:
So, this means that in call queries we need to add `startLine = sourcecode.Line()`, right? Would it be possible to introduce this as a default argument?

PR #1096, opened 2021-01-22
Integrate with scalafix for OrderingImports

hubertp:
Initial integration solely for the purpose of solving the unordered
imports problem. scalafix can introduce a lot of other linting, but for
now I'm only invoking a single rule.

Provides `format` alias to correctly first invoke scalafix and then
scalafmt (the former can break formatting). Having a single command
simplifies stuff when we have failures in CI.

Not used TaskKey definition as sbt-scalafix does some "clever" caching
preventing me from invoking it correctly. Alias works just fine.

Note:
This change does not introduce PR/master checks for scalafixCheck.
This change does not include the result of `sbt format`.

Both will be added in a follow up PR, as they do a lot of automatic
formatting.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -11,11 +11,16 @@ inThisBuild(
       Resolver.mavenLocal,
       Resolver.bintrayRepo("shiftleft", "maven"),
       Resolver.bintrayRepo("mpollmeier", "maven"),
-      "Sonatype OSS" at "https://oss.sonatype.org/content/repositories/public"),
+      "Sonatype OSS" at "https://oss.sonatype.org/content/repositories/public"
+    ),
     packageDoc / publishArtifact := true,
     packageSrc / publishArtifact := true,
-    scmInfo := Some(ScmInfo(url("https://github.com/ShiftLeftSecurity/codepropertygraph"),
-                            "scm:git@github.com:ShiftLeftSecurity/codepropertygraph.git")),
+    scmInfo := Some(
+      ScmInfo(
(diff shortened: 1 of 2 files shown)


Ferada:
And you're sure you want to do this via `sbt` and not pre-commit via command-line version?

hubertp:
In this repo we already do scalafmtCheck as part of the build process. 
Judging from some of the past PRs that seems to be the preferred way, as I often see `scalafmt` commits, meaning it is done manually.

Ferada (approved):

fabsx00:
So, for the user, what does that mean now? What needs to be run before pushing? Still `sbt scalafmt` and `sbt test:scalafmt` or `sbt format`?

build.sbt
```diff
@@ -42,16 +47,21 @@ inThisBuild(
...
 
-ThisBuild/libraryDependencies ++= Seq(
-  "org.apache.logging.log4j" %  "log4j-slf4j-impl" % "2.11.2" % Test
+ThisBuild / libraryDependencies ++= Seq(
+  "org.apache.logging.log4j" % "log4j-slf4j-impl" % "2.11.2" % Test
 )
 
+// Scalafix / imports check setup
+ThisBuild / scalafixDependencies += "com.github.liancheng" %% "organize-imports" % "0.5.0-alpha.1"
+ThisBuild / semanticdbEnabled := true // enable SemanticDB
```
mpollmeier, line 62: that comment doesn't really add anything

mpollmeier (approved):

fabsx00 (approved):

PR #1099, opened 2021-01-28
make queryInit explicitly return a `Query` to avoid casts

mpollmeier:
note: in order to reference `Query` from the macro i needed to pull it
out to another common dependency - you probably want to move it
somewhere else


ursachec (approved):
ah, nice. didn't think of moving it to a separate ~namespace~ subproject

mpollmeier:
`s/namespace/subproject/` :)

fabsx00 (approved):

mpollmeier:
@ursachec i assumed you wanted to move the query class somewhere else, or did you want to keep it where it is? I don't mind either way

ursachec:
@mpollmeier looks ok as it is at the moment. we can move it later if necessary

PR #1103, opened 2021-01-31
Set graphviz direction explicitly

AlexDenisov:
Before:

ast

After:

ast-lr

@fabsx00 please let me know if there is a specific reason to not set the direction explicitly.


fabsx00 (commented):
No specific reason, although, intuitively, I'd think the tree has its root at the top and children are at the bottom. The left-to-right version seems more practical though, you're right. Keep in mind though that if you patch it here, it will also influence CFG and PDG plotting. Overall, all the structures would be rotated by 90 degrees compared to the way they typically show up in the literature.

AlexDenisov:
Right, I just looked at the ddg/pdg and the graphs look good in the default top-to-bottom layout.
I guess there is no single default that would nicely fit all cases. I'll try making this configurable then and update the PR soon.

ursachec:
@AlexDenisov stale, closing lovely

PR #1106, opened 2021-02-01
More incremental work on CPG spec cleanup

fabsx00:
* Cleanup and extend File, MetaData, Method and NamespaceBlock tests
* Use `typeFullName` in place of `evalType` where immediately possible
* Move `AST_PARENT_TYPE` and `AST_PARENT_FULL_NAME` into enhancements

### DIFF
--- a/fuzzyc2cpg-tests/src/test/scala/io/shiftleft/fuzzyc2cpg/standard/MethodTests.scala
+++ b/fuzzyc2cpg-tests/src/test/scala/io/shiftleft/fuzzyc2cpg/standard/MethodTests.scala
@@ -7,8 +7,8 @@ class MethodTests extends FuzzyCCodeToCpgSuite {
 
   override val code =
     """
-      |int main(int argc, char **argv) {
-      |}""".stripMargin
+      |  int main(int argc, char **argv) {
+      | }""".stripMargin
 
   "should contain exactly one method node with correct fields" in {
     cpg.method.l match {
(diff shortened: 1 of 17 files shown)


tuxology (approved):

fuzzyc2cpg-tests/src/test/scala/io/shiftleft/fuzzyc2cpg/standard/MethodTests.scala
```diff
@@ -3,60 +3,48 @@ package io.shiftleft.fuzzyc2cpg.standard
...
-    cpg.method.lineNumber.l shouldBe List(2)
-  }
-
-  "should return correct end line number" in {
-    cpg.method.lineNumberEnd.l shouldBe List(3)
+      |  int main(int argc, char **argv) {
+      | }""".stripMargin
+
+  "should contain exactly one method node with correct fields" in {
+    cpg.method.l match {
```
mpollmeier, line 14: i noticed this style is used in many places... wouldn't the following be easier to read? 
```
val List(x) = cpg.method.l
x.name shouldBe "main"
...
```

Semantically it's the same, i.e. it'll fail if there's zero or >1 elements in the list. 
I'm really not sure, it may be more confusing to a scala newbie, but to me it looks cleaner, and is shorter. Importantly, it doesn't have any indentation - I've seen nested versions of this pattern too.

mpollmeier (approved):

fuzzyc2cpg-tests/src/test/scala/io/shiftleft/fuzzyc2cpg/standard/MethodTests.scala
fabsx00: oh, interesting, I didn't know this was possible

prabhu (approved):

fuzzyc2cpg-tests/src/test/scala/io/shiftleft/fuzzyc2cpg/standard/MethodTests.scala
fabsx00: Changed it for the tests in `standard` and will change it in other places I discover in the future.

prabhu (approved):

PR #1150, opened 2021-03-16
Add `language` parameter to Query case class

ursachec:
And set it automatically based on the namespace it finds itself in.

### DIFF
--- a/console/src/main/scala/io/shiftleft/console/QueryDatabase.scala
+++ b/console/src/main/scala/io/shiftleft/console/QueryDatabase.scala
@@ -45,8 +45,13 @@ class QueryDatabase(defaultArgumentProvider: DefaultArgumentProvider = new Defau
     * */
   def queriesInBundle[T <: QueryBundle](bundle: Class[T]): List[Query] = {
     queryCreatorsInBundle(bundle).map {
-      case (method, args) =>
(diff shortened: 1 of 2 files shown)


itsacoderepo (approved):

console/src/main/scala/io/shiftleft/console/QueryDatabase.scala
```diff
@@ -45,8 +45,13 @@ class QueryDatabase(defaultArgumentProvider: DefaultArgumentProvider = new Defau
     * */
   def queriesInBundle[T <: QueryBundle](bundle: Class[T]): List[Query] = {
     queryCreatorsInBundle(bundle).map {
-      case (method, args) =>
-        method.apply(args: _*).asInstanceOf[Query]
+      case (method, args) => {
+        val query = method.apply(args: _*).asInstanceOf[Query]
+        val bundleNamespace = classToType(bundle).typeSymbol.fullName.toString
+        // the namespace currently looks like `io.joern.scanners.c.CopyLoops`
+        val language = bundleNamespace.split('.')(3)
```
fabsx00, line 52: Does this `(3)` throw an exception if the namespace has fewer .-separated elements? If so, maybe we can implement a version here that tries to fetch the fourth element but returns an empty string if that fails.

PR #1154, opened 2021-03-24
fix intellij / bsp build: remove the baseDirectory setting

mpollmeier:
This innocent-looking setting made intellij choke on the subprojects and
didn't recognise them as scala projects properly. Unfortunately this
setting was necessary because intellij and sbt have different default
working directories for executing tests. 

This PR uses a hack to achieve the same: it looks for a file that (hopefully always)
only be in the project root. If the current working dir doesn't contain that file,
we must be in a subproject, and will therefor navigate up one directory.
Turns out: this same technique was already applied in ScriptManagerTest.scala before - someone else has a knack for hacks

And just to motivate why BSP is a good idea: it ensures that Intellij and sbt share the compiled classes etc., so we don't need to compile twice, with potentially different results. Initial import is also much faster. 

All tests from within Intellij seem to be green.

### DIFF
--- a/codepropertygraph/src/main/scala/io/shiftleft/utils/ProjectRoot.scala
+++ b/codepropertygraph/src/main/scala/io/shiftleft/utils/ProjectRoot.scala
@@ -1,0 +1,30 @@
+package io.shiftleft.utils
+
+import better.files.File
+
+/**
+  * Finds the relative location of the project root.
+  *
+  * Used in tests which rely on the working directory - unfortunately Intellij and sbt have different default working
+  * directories for executing tests from subprojects: while sbt defaults to the project root, intellij defaults to
+  * the subproject.
+  *
+  * Previously a consistent behaviour was achieved by setting
+  * `Test / baseDirectory := (ThisBuild / Test / run / baseDirectory).value`,
+  * however that broke the bsp build within Intellij - it simply wouldn't recognise subprojects with this setting
+  * any more.
+  */
+object ProjectRoot {
+
+  def relativise(path: String): String =
+    s"$findRelativePath/$path"
+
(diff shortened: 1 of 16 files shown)


mpollmeier:
If this finds happy reviewers I'll apply the same changes to the policy and codescience repositories, update the wiki and let the team know.

Ferada (approved):
Impressive work! I can confirm that after reimporting via BSP I do not see any immediate build problems and all tests both seem to be recognised as such and can be run without problems. I get a few errors, but that's probably because I just ran with a Java 14 JDK instead of 8. For me :+1:

codepropertygraph/src/main/scala/io/shiftleft/utils/ProjectRoot.scala
```diff
@@ -0,0 +1,30 @@
...
+
+  def relativise(path: String): String =
+    s"$findRelativePath/$path"
+
+  def findRelativePath: String = {
+    val fileThatOnlyExistsInRoot = ".git"
+
+    if (File(fileThatOnlyExistsInRoot).exists) "."
+    else if (File(s"../$fileThatOnlyExistsInRoot").exists) ".."
+    else ???
```
hubertp, line 27: I'd prefer we don't use `???` if finding fails. `None` sounds better

hubertp (commented):
Nit about `???`

codepropertygraph/src/main/scala/io/shiftleft/utils/ProjectRoot.scala
mpollmeier: Hmm, I'd prefer to keep the ??? because this can easily be filled if that use case ever arises..

PR #1185, opened 2021-04-14
rip protoBindings

mpollmeier:
* some are being built in internal builds
* anyone can build their own based on the cpg.proto which is still
  being published in the schema.jar


fabsx00 (approved):

mpollmeier:
this turns out to be more complicated than I thought, mostly because the CpgOverlay depends on the cpg proto bindings.

PR #1187, opened 2021-04-22
don't handle schema extensions in plugin manager any more

mpollmeier:
This one may be controversial, so let me explain:
We used to define the schema in json, which meant we didn't have any
IDE and compiler support. Now that it's a Scala DSL we're in a much
better situation overall, however we cannot simply copy and merge
schema parts together by copying a json file. The schema needs to be
stitched together at compile time.

Now we could allow plugins to define their schema extension parts in
scala code snippets and copy them together during `./joern
--add-plugin` - this is pretty much what I wanted to achieve in
https://github.com/joernio/docker-ext/pull/9/ - the downside is that
we lose IDE/compiler support during plugin development...

tl;dr; there's different options/pros/cons, and we may find a better
model in the future, but for now I think this is the best compromise
with the new improved schema DSL.

I should add: the responsibility to create new domain classes and install them into joern now lies with the plugin itself - we can extract that logic into an sbt plugin later.

### DIFF
--- a/console/src/main/scala/io/shiftleft/console/PluginManager.scala
+++ b/console/src/main/scala/io/shiftleft/console/PluginManager.scala
@@ -77,10 +77,6 @@ class PluginManager(val installDir: File) {
         dir.list.filter { f =>
           f.name.startsWith(s"joernext-$name")
         }
-      }.flatMap { dir =>
-        dir.list.filter { f =>
-          f.name.startsWith(s"joernext-$name")
-        }
       }
       filesToRemove.foreach(f => f.delete())
       filesToRemove.map(_.pathAsString)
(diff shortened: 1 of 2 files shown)


fabsx00:
So, can a plugin still add the schema without having joern/ocular know the plugin?

mpollmeier:
If I understand your question correctly: yes. To be safe, let me rephrase the main points:

* the joern/ocular repos/builds don't have any reference to the plugin which extends the schema - I guess that's your main question?
* if a plugin extends the schema, it contains a bit of boilerplate build definition to install the schema extension into a local joern/ocular installation
* that boilerplate can eventually move into an sbt plugin to keep the plugin build concise

Makes sense?

fabsx00 (approved):
That seems fine to me then. Of course, a bit more boilerplate in the plugin isn't a nice thing to have, but we gain the goodies of developing the schema in the IDE, and that seems worth it. As long as joern/ocular don't need to somehow know of the plugin before hand, we're good, so, yes, this looks good to me.

PR #1189, opened 2021-04-26
fix all warnings and make warnings errors

mpollmeier:


console/src/main/scala/io/shiftleft/console/BridgeBase.scala
```diff
@@ -280,7 +280,7 @@ trait BridgeBase {
       try {
         websiteContentDir.delete()
       } catch {
-        case e: IOException => {
+        case _: IOException => {
```
hubertp, line 283: Should probably include stacktrace

hubertp (approved):

console/src/main/scala/io/shiftleft/console/BridgeBase.scala
mpollmeier: I wanted to keep this to refactoring only - this is for the ammonite bridge which is used for CpgServer as well, so it could well be that we want to be conservative when it comes to printing stuff... but I'm not sure really

PR #1190, opened 2021-04-27
sbt 1.5.1

mpollmeier:


hubertp (approved):
i couldn't see how sbt is fetched in GH actions but other than that lgtm

mpollmeier:
it's downloading the small sbt wrapper script, so I presume it'll work... https://github.com/olafurpg/setup-scala/blob/master/src/install.ts#L105

PR #1192, opened 2021-04-27
Bring back in TYPE_FULL_NAME on CALL nodes.

ml86:
This was once accedidentially put to deprecated and in the resent past
remove while transitioning to scala base schema.


fabsx00:
While we're at it, what is the TYPE_FULL_NAME of a call? The type of its return value?

fabsx00 (approved):

PR #1213, opened 2021-05-16
persist codegen/protogen caches, to keep it between sbt runs

mpollmeier:


schema/build.sbt
```diff
@@ -2,22 +2,22 @@ name := "codepropertygraph-schema"
 
 libraryDependencies += "io.shiftleft" %% "overflowdb-codegen" % "1.63"
 
-val generateDomainClasses = taskKey[Seq[File]]("generate overflowdb domain classes for our schema")
+scalacOptions += "-Xfatal-warnings"
```
ml86, line 5: Is that intended?

ml86 (approved):

schema/build.sbt
mpollmeier: yup, i just moved that line within the same file

PR #1220, opened 2021-05-26
WIP - return Traversal[CallRepr] instead of Traversal[Call]

mpollmeier:
this may contain ImplicitCalls
re https://github.com/ShiftLeftSecurity/product/issues/7632


mpollmeier:
will go with https://github.com/ShiftLeftSecurity/codepropertygraph/pull/1223 instead

PR #1223, opened 2021-05-26
fix classCast exceptions for implicit calls in callgraphExtensions

bbrehm:
Per interface, the call resolver returns `CallRepr`; however, we simply cast this to `Call` in `someMethod.call`. It is possible to get other `CallRepr` here, namely `ImplicitCall` and `PostExecutionCall`. These then lead to a ClassCastException.

This interface mismatch is somewhat unfortunate. The immediate way forward is to just silently swallow other callRepr here, because typical users (ourselves in framework passes and scripts) of the extension methods are unprepared to deal with other CallRepr. Sufficiently sophisticated users (ourselves when working on core engine stuff) who want to deal with the fancy CallRepr subtypes can use the direct API of the CallResolver.

This should fix https://github.com/ShiftLeftSecurity/product/issues/7632 and should obsolete https://github.com/ShiftLeftSecurity/codescience/pull/5017 (but kudos to @mpollmeier for the super fast workaround!)

This is an alternative to https://github.com/ShiftLeftSecurity/codepropertygraph/pull/1220

### DIFF
--- a/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/callgraphextension/Method.scala
+++ b/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/callgraphextension/Method.scala
@@ -46,7 +46,7 @@ class Method(val traversal: Traversal[nodes.Method]) extends AnyVal {
     * Incoming call sites
     * */
   def callIn(implicit callResolver: ICallResolver): Traversal[nodes.Call] =
-    traversal.flatMap(method => callResolver.getMethodCallsitesAsTraversal(method).cast[nodes.Call])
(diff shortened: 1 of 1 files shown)


ml86:
Lets take this fix. Implementing a version which returns CallRepr is would mean exposing the user to ImplicitCall and PostExecutionCall which we do not want.

ml86 (approved):

mpollmeier:
You guys are deeper in the cpg spec and I'm fully onboard if you agree on this. 
I just wanted to note that if we go for this option (i.e. filter out anything other than `Call`), we may be missing out on flows that include an ImplicitCall. Unless we add those elsewhere, e.g. as you noted in the description "use direct API of the CallResolver" in TrackingPointExt

PR #1236, opened 2021-06-01
cache hashes of accesspaths; keep exclusions of accesspaths sorted

bbrehm:
The caching of accesspath hashes gives > 5% speedup on dataflow heavy loads (cumulative to https://github.com/ShiftLeftSecurity/overflowdb/pull/194).

It is also saves some expansions (more context merges) in the closed source tracker; and it is a useful preparation for OVERRIDE/PRESERVE handling of policies.

### DIFF
--- a/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/accesspath/AccessPath.scala
+++ b/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/accesspath/AccessPath.scala
@@ -24,13 +24,15 @@ object AccessPath {
   private implicit class ElementsDecorations(el: Elements) {
 
     def noOvertaint(start: Int = 0, untilExclusive: Int = el.elements.length): Boolean = {
-      el.elements
-        .slice(start, untilExclusive)
-        .find {
-          case (VariablePointerShift | VariableAccess) => true
-          case _                                       => false
+      var idx = start
+      while (idx < untilExclusive) {
+        el.elements(idx) match {
+          case (VariablePointerShift | VariableAccess) => return false
+          case _                                       =>
         }
-        .forall(_ => false)
+        idx += 1
+      }
+      true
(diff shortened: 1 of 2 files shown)


semanticcpg/src/main/scala/io/shiftleft/semanticcpg/accesspath/AccessPath.scala
```diff
@@ -49,9 +51,10 @@ object AccessPath {
       val nElements = el.elements.length - 1
       while (nElements - i > -1) {
         el.elements(nElements - i) match {
-          case (AddressOf | VariablePointerShift | _: PointerShift) => i += 1
+          case (AddressOf | VariablePointerShift | _: PointerShift) =>
           case _                                                    => return i
         }
+        i += 1
```
ml86, line 57: Why?

ml86 (approved):

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/accesspath/AccessPath.scala
fabsx00: Do we need this unaesthetic while loop here?

bbrehm: This specific change is purely for readability -- it does not change the control flow graph, and if the scala compiler does its job then it should not even change the bytecode.

Regarding whether we need the loop at all -- this is hot code, and I see no formulation that is both more obvious in terms of semantics and more obvious in terms of performance (i.e. in terms of what we hope for the style of machine code).

ml86: To me the old version was the better one.

Issue #1233, opened 2021-06-01
RFC: rename semanticcpg.language.types.expressions.Call -> CallExt etc. 


mpollmeier:
I'd like to rename all traversal extension classes like e.g.  `semanticcpg.language.types.expressions.Call` to `semanticcpg.language.types.expressions.CallTrav`. 

Motivation: that enables us to drop the `nodes.` prefix everywhere. It somehow became a norm in our codebase to refer to node types with the explict prefix, e.g. `nodes.Call`. The more I see that pattern the less I like it, and yes I started it, but I feel sorry about it. 

Before I go ahead and create a monster PR I thought I'd ask for feedback. 
Please select one of the following options: yes|no|dontcare

ml86:
I am all for it.I already suggest it 1-2 years back but did not come through with it. So hoping for the best now.

mpollmeier:
Just tried to do the same for the codescience repo, but unfortunately things are more complicated there...
Shadowing implicits in generated code and cpg repo, yet again a different naming convention make it quite complicated to clean up and standardise the codescience repo. 
I'll put this on hold until we've migrated to Scala3, where extension methods don't need to be in (named) classes any more.

PR #1253, opened 2021-06-11
refactor: standardise extension classes, drop `nodes.` prefix

mpollmeier:
re https://github.com/ShiftLeftSecurity/codepropertygraph/issues/1233


fabsx00 (approved):
I guess we should also introduce a convention for the class names of classes that wrap and extend nodes, e.g., `MethodMethods`. Currently, the convention seems to be $NameMethods.

mpollmeier:
yes that seems to be our convention, and I think we're fairly consistent. 
we'll soon not need a convention for those extension methods any more, since they don't need to have a name any more in scala 3: https://docs.scala-lang.org/scala3/book/ca-extension-methods.html

PR #1256, opened 2021-06-13
Refactor LocationCreator

fabsx00:


semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/nodemethods/AstNodeMethods.scala
```diff
@@ -2,9 +2,10 @@ package io.shiftleft.semanticcpg.language.nodemethods
 
 import io.shiftleft.Implicits.JavaIteratorDeco
 import io.shiftleft.codepropertygraph.generated.nodes._
+import io.shiftleft.semanticcpg.NodeExtension
 import io.shiftleft.semanticcpg.language._
 
-class AstNodeMethods(val node: AstNode) extends AnyVal {
+class AstNodeMethods(val node: AstNode) extends NodeExtension {
```
mpollmeier, line 8: Using a mixin trait here unfortunately has the consequence that these extension classes are no longer value classes
Value classes have the benefit of non-existence at runtime - they provide compile time guarantees without runtime cost. 
https://docs.scala-lang.org/overviews/core/value-classes.html

Given that `NodeExtensionFinder` instantiates it, wraps it in an Option and returns it, this may simply be the price to pay for this design.

mpollmeier (approved):

fabsx00:
As discussed on slack, `extends Any` introduces a universal trait and with that, we should be allocation-free here until someone makes use of whatever the universal trait defines.

PR #1259, opened 2021-06-15
downgrade ammonite - when using 2.4.0 we get scala3 macro related errors in cs

mpollmeier:
```
Unsupported Scala 3 inline macro method generate; found in trait
sourcecode.NameMacros.
```


mpollmeier:
found a better workaround

PR #1275, opened 2021-06-25
Language primitives for control structure traversal

fabsx00:
We recently introduced the `controlStructureType` field but never got around to extending the query language accordingly, e.g., to allow traversing to all if-blocks or all while-blocks. This PR adds the necessary language primitives as well as a querying test.


itsacoderepo (approved):

mpollmeier:
:100:

PR #1276, opened 2021-06-28
upgrade odb codgen

mpollmeier:
Note the change in TypeDeclTests - this is an intentional change in the
codegen, it used to contain a special handling for integers, which 
defaulted to -1. Please let me know if this breaks any other assumptions
down the line... Looks to me like `order` should be an optional property.


bbrehm:
What is the status with respect to default values? We should wait for that with this change, right?

mpollmeier:
that's what i first thought as well, but I'll split up the two, i.e. I'll bring back the implicit default value of -1 for integers in the codegen pr

PR #1281, opened 2021-07-06
Exit gracefully on CPGQLS startup exceptions

ursachec:
Without this change, an exception encountered during server startup
causes the shell session to be blocked, not accepting any input.

Example message:
```
$ ./joern --server
Starting CPGQL server ...
Unhandled exception thrown while attempting to start CPGQL server, exiting.
```

### DIFF
--- a/console/src/main/scala/io/shiftleft/console/BridgeBase.scala
+++ b/console/src/main/scala/io/shiftleft/console/BridgeBase.scala
@@ -308,7 +308,20 @@ trait BridgeBase {
                                  config.serverAuthUsername,
                                  config.serverAuthPassword)
     println("Starting CPGQL server ...")
-    server.main(Array.empty)
+    try {
+      server.main(Array.empty)
+    } catch {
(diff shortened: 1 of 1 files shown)


maltek (approved):

console/src/main/scala/io/shiftleft/console/BridgeBase.scala
```diff
@@ -308,7 +308,20 @@ trait BridgeBase {
...
+    try {
+      server.main(Array.empty)
+    } catch {
+      case _: java.net.BindException => {
+        println("Could not bind socket for CPGQL server, exiting.")
+        ammonite.shutdown()
+        System.exit(1)
+      }
+      case _: Throwable => {
+        println("Unhandled exception thrown while attempting to start CPGQL server, exiting.")
```
maltek, line 320: maybe print the message here?

ursachec: Yes, good point

PR #1282, opened 2021-07-07
Close connections gracefully in CPGQL server

ursachec:
Previously the server did not dispose of closed connections and instead
tried to send messages to them leading to exceptions.

### DIFF
--- a/console/src/main/scala/io/shiftleft/console/cpgqlserver/CPGQLServer.scala
+++ b/console/src/main/scala/io/shiftleft/console/cpgqlserver/CPGQLServer.scala
@@ -57,7 +57,18 @@ class CPGQLServer(ammonite: EmbeddedAmmonite,
       connection.send(cask.Ws.Text("connected"))
       openConnections += connection
       cask.WsActor {
-        case cask.Ws.Close(_, _) => openConnections -= connection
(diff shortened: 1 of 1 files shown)


console/src/main/scala/io/shiftleft/console/cpgqlserver/CPGQLServer.scala
```diff
@@ -57,7 +57,18 @@ class CPGQLServer(ammonite: EmbeddedAmmonite,
       connection.send(cask.Ws.Text("connected"))
       openConnections += connection
       cask.WsActor {
-        case cask.Ws.Close(_, _) => openConnections -= connection
+        case cask.Ws.ChannelClosed() => {
```
maltek, line 60: doesn't really matter but could merge the two cases `case cask.Ws.ChannelClosed() | cask.Ws.Close(_, _) => {`

maltek (approved):

console/src/main/scala/io/shiftleft/console/cpgqlserver/CPGQLServer.scala
ursachec: they do look nicer like that, thx!

PR #1284, opened 2021-07-07
add ghidra frontend

mpollmeier:


mpollmeier:
not really sure about the naming here, since it kinda violates our convention of using the programming language name as the frontend name. not sure if there's a better way, since ghidra is somewhat different...

fabsx00 (approved):

fabsx00:
Let's stay with the naming for now as we're sure to have an ida2cpg at some point, and then both would be "binary2cpg".

itsacoderepo (approved):

PR #1290, opened 2021-07-13
Add varArgParameterIndex property to METHOD node.

ml86:
See schema comment.

### DIFF
--- a/schema/src/main/scala/io/shiftleft/codepropertygraph/schema/Method.scala
+++ b/schema/src/main/scala/io/shiftleft/codepropertygraph/schema/Method.scala
@@ -72,6 +72,19 @@ object Method extends SchemaBase {
       .addProperties(astParentType, astParentFullName)
       .extendz(declaration)
 
+    val isVariadic = builder
+      .addProperty(
+        name = "IS_VARIADIC",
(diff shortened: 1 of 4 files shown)


schema/src/main/scala/io/shiftleft/codepropertygraph/schema/Method.scala
```diff
@@ -40,6 +40,21 @@ object Method extends SchemaBase {
...
       .protoId(22)
 
+    val varArgParameterIndex = builder
+      .addProperty(
+        name = "VARARG_PARAMETER",
+        valueType = ValueTypes.INTEGER,
+        cardinality = Cardinality.One,
+        comment =
+          """
+            |For variadic methods the VARARG_PARAMETER property specifies which parameter
```
fabsx00, line 50: This is a bit weird because it means that the `METHOD` node stores information about another node, namely, the parameter. Any reason why we don't have an `IS_VARIADIC` field in the parameter instead?

fabsx00 (changes requested):
see comment.

ml86:
As discussed in zoom, i introduced IS_VARIADIC property.
Please have another look.

fabsx00 (approved):

PR #1293, opened 2021-07-14
Change operator extensions

bbrehm:
The old implementation of operator extensions subclasses Call, and then allocates a new subclassed Call that wraps the same underlying CallDb object with the same id.

Unfortunately, this is breaks underlying assumptions of overflowdb: It is one of the main design constraints of overflowdb that there exists one and only one NodeRef (the superclass of Call) for every node (`public final Noderef ref;` in `NodeDb`). Creating a second one does not work and leads to undebuggable corruption:

-The noderef initially does not get registered.
-eventually, the canonical ref to the underlying nodeDb gets cleared, and the nodeDb gets written to disk
-the canonical ref gets reloaded. Now we have two NodeDb instances with the same id.
-one of them gets modified, and the modifications are invisible to the second one.
-No idea what happens when we try to serialize the entire thing again.

I think we should generally get rid of the OperatorExtensions. They are used very rarely. In lieu of complete removal, this PR changes the API to a supportable one.

cc @mpollmeier for whether my view on the validity of this old construction is correct.
cc @fabsx00 for the new (much uglier) API

If we merge this, then I'll do a second round in joern. Same if we decide to just kick out OperatorExtensions.

### DIFF
--- a/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/operatorextension/Nodes.scala
+++ b/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/operatorextension/Nodes.scala
@@ -3,10 +3,7 @@ package io.shiftleft.semanticcpg.language.operatorextension
 import io.shiftleft.codepropertygraph.generated.nodes.Call
 
 object opnodes {
-  case class Assignment(val call: Call) // extends Call(call.graph, call.id)
-  case class Arithmetic(val call: Call) // extends Call(call.graph, call.id)
-  case class ArrayAccess(val call: Call) // extends Call(call.graph, call.id)
-  implicit def unwrap(assignment: Assignment): Call = assignment.call
-  implicit def unwrap(assignment: Arithmetic): Call = assignment.call
-  implicit def unwrap(assignment: ArrayAccess): Call = assignment.call
+  case class Assignment(val call: Call)
+  case class Arithmetic(val call: Call)
+  case class ArrayAccess(val call: Call)
 }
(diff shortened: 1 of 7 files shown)


fabsx00 (changes requested):
It isn't true that operator extension isn't used much. We use it heavily for interactive querying in the C world. The fact that it's not used internally is probably mostly due to the fact that this extension was introduced late. The idea here is to provide a decorator language, that is , a language that introduces types that are not as corresponding CPG node types. As far as the adaptions you make in this PR are concerned, they're lengthy and clumsy, so, I don't think we should merge this.

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/operatorextension/Nodes.scala
```diff
@@ -3,7 +3,7 @@ package io.shiftleft.semanticcpg.language.operatorextension
 import io.shiftleft.codepropertygraph.generated.nodes.Call
 
 object opnodes {
-  class Assignment(call: Call) extends Call(call.graph, call.id)
-  class Arithmetic(call: Call) extends Call(call.graph, call.id)
-  class ArrayAccess(call: Call) extends Call(call.graph, call.id)
+  case class Assignment(val call: Call)
```
mpollmeier, line 6:
```suggestion
  case class Assignment(call: Call)
```
case class members are publicly accessible by default, same for the lines below

mpollmeier (commented):
I guess we all agree that the new API is a step backwards. We can probably introduce some implicits to make it look similar to before, but that may be more effort than fixing the actual issue here, which is a (rarely if ever triggered, but existent) bug in overflowdb.

PR #1296, opened 2021-07-15
use jdk11 for main build, but use source/target=1.8

mpollmeier:
idea: we want to be able to use these artifacts in downstream builds that require jdk8

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -17,10 +17,11 @@ jobs:
       run: git submodule update --init --recursive
     - name: apt-get update
       run: sudo apt-get update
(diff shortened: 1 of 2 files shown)


fabsx00 (approved):

max-leuthaeuser:
PR builds currently using a build matrix (8 and 11). This needs to be adapted as well doesn't it?

mpollmeier:
I saw that, but thought we'll leave it in until you bring in the changes that don't work on JDK8, also to make this change visible. 
I don't mind either way, we can remove the JDK8 PR build now if you like.

PR #1310, opened 2021-07-26
Proposal for new CpgPass interface.

ml86:
Comments/docu is still old, so dont get confused.


bbrehm:
As per zoom discussion, we need to think a little longer about scheduling and guarantees.

bbrehm (changes requested):
As discussed on slack/zoom, we're not done yet, and leave this on ice for some time.

Hitting the "Request changes" button to mute github reminders; feel free to re-request review once we're ready to move forward.

fabsx00:
I guess this one can be closed since we merged the alternative proposal.

PR #1318, opened 2021-07-30
improve diffGraphBuilder API

bbrehm:
1. allow fluent/builder pattern, i.e. return `this` on changes. 
2. there is no need for diffgraph builders to remember what new nodes they contain (this is afaiu vestigal code from before we fixed containedNode performance).
3. add utility functions to make parallel mutable reductions more efficient and convenient: That is, convenience functions for effectively merging diffgraphs.

### DIFF
--- a/codepropertygraph/src/main/scala/io/shiftleft/passes/DiffGraph.scala
+++ b/codepropertygraph/src/main/scala/io/shiftleft/passes/DiffGraph.scala
@@ -192,11 +192,10 @@ object DiffGraph {
   def newBuilder: Builder = new Builder()
 
   class Builder {
-    private var _buffer: mutable.ArrayBuffer[Change] = null
-    private var _nodeSet: java.util.IdentityHashMap[NewNode, NewNode] = null
-    private def buffer: mutable.ArrayBuffer[Change] = {
+    private var _buffer: ChangeBuffer = null
+    private def buffer: ChangeBuffer = {
       if (_buffer == null)
-        _buffer = new mutable.ArrayBuffer[Change]()
+        _buffer = new ChangeBuffer()
       _buffer
     }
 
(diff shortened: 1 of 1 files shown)


codepropertygraph/src/main/scala/io/shiftleft/passes/DiffGraph.scala
```diff
@@ -220,42 +220,50 @@ object DiffGraph {
     def build(): DiffGraph = build(_buffer)
     def buildReverse(): DiffGraph = build(if (_buffer != null) _buffer.reverse else null)
 
-    def addNode(node: NewNode): Boolean = {
-      if (_nodeSet == null) _nodeSet = new java.util.IdentityHashMap[NewNode, NewNode]()
```
mpollmeier, line 224: you're sure we don't need the IdentityHashMap any more? 
I guess it's fine because the generated NewNode instances implement equals and that looks ok...
```
   override def equals(other: Any): Boolean = other match {
      case that: NewBlockBuilder => (that canEqual this) && _id == that._id
      case _ => false
   }
```

mpollmeier (approved):

codepropertygraph/src/main/scala/io/shiftleft/passes/DiffGraph.scala
bbrehm: It's completely unrelated.

The NewNode equality is highly misleading; for all intents and purposes, NewNode should compare by identity, and not be a case class. If I have two NewTag nodes A and B with `(name="foo", value="bar")` (compare equal, but different identity) and I add edges from A to method "Foo" and an edge from B to method "Bar", then I expect two different StoredNode to get serialized, with the correct edges.

I have argued against this again and again, but somehow NewNode becomes a case class with overridden `equals` again and again.

mpollmeier: again and again? They've been case classes ever since the initial commit in the odb codegen repo (april 2020), with no changes afaics.

bbrehm: Apologies, my tone was uncalled for. Some day we may fix it; but it's no big problem.

The specific identityHashMap here is just unneeded; I vaguely remember that it was used back when the diffgraph builder was responsible for adding transitive NewNodes (i.e. NewNodes that have not been explicitly added to the builder, but are referenced by elements that have been added). Nowadays, the DiffgraphApplier is responsible for dealing with that (and it indeed has an IdentityHashMap for it).

mpollmeier: :+1:

ml86 (approved):

PR #1317, opened 2021-07-30
Clarify diffgraph application, improve log messages, remove createApplyAndSerialize

bbrehm:
This does not meaningfully change CpgPass semantics or API; it simply documents the current situation.

PR #1319, opened 2021-07-30
Alternative proposal for CpgPass interface

bbrehm:
Related https://github.com/ShiftLeftSecurity/codepropertygraph/pull/1310 https://github.com/ShiftLeftSecurity/codepropertygraph/pull/1317

### DIFF
--- a/codepropertygraph/src/main/scala/io/shiftleft/passes/CpgPass.scala
+++ b/codepropertygraph/src/main/scala/io/shiftleft/passes/CpgPass.scala
@@ -7,6 +7,7 @@ import io.shiftleft.codepropertygraph.generated.nodes.{NewNode, StoredNode}
 import org.slf4j.{Logger, LoggerFactory, MDC}
 
 import java.lang.{Long => JLong}
+import java.util.function.{BiConsumer, Supplier}
 import scala.concurrent.duration.DurationLong
 
 /**
@@ -83,9 +84,88 @@ abstract class CpgPass(cpg: Cpg, outName: String = "", keyPool: Option[KeyPool] 
 
 }
 
+abstract class NewStylePass[T <: AnyRef](cpg: Cpg, outName: String = "", keyPool: Option[KeyPool] = None)
+    extends CpgPassBase {
+  //generate Array of parts that can be processed in parallel
+  def generateParts(): Array[_ <: AnyRef] = Array[AnyRef](null)
+  //setup large data structures, acquire external resources
+  def init(): Unit = {}
+  //release large data structures and external resources
+  def finish(): Unit = {}
+  //main function: add desired changes to builder
+  def runOnPart(builder: DiffGraph.Builder, part: T): Unit
+
+  override def createAndApply(): Unit = createApplySerializeAndStore(null)
+
+  override def createApplySerializeAndStore(serializedCpg: SerializedCpg,
+                                            inverse: Boolean = false,
+                                            prefix: String = ""): Unit = {
+    baseLogger.info(s"Start of enhancement: $name")
+    val tic = System.nanoTime()
+    var nParts = 0
+    var tocBuild = -1L
+    var size = -1
+    try {
+      init()
+      val parts = generateParts()
+      nParts = parts.size
+      val diffGraph = nParts match {
+        case 0 => DiffGraph.EmptyChangeSet
+        case 1 =>
+          val builder = DiffGraph.newBuilder
+          runOnPart(builder, parts(0).asInstanceOf[T])
+          builder.build()
+        case _ =>
+          java.util.Arrays
+            .stream(parts)
+            .parallel()
+            .collect(
+              new Supplier[DiffGraph.Builder] {
+                override def get(): DiffGraph.Builder = DiffGraph.newBuilder
+              },
+              new BiConsumer[DiffGraph.Builder, AnyRef] {
+                override def accept(builder: DiffGraph.Builder, part: AnyRef): Unit =
+                  runOnPart(builder, part.asInstanceOf[T])
+              },
+              new BiConsumer[DiffGraph.Builder, DiffGraph.Builder] {
+                override def accept(leftBuilder: DiffGraph.Builder, rightBuilder: DiffGraph.Builder): Unit =
+                  leftBuilder.addChanges(rightBuilder)
+              }
+            )
+            .build()
+      }
+      tocBuild = System.nanoTime()
+      size = diffGraph.size
+      val withInverse = serializedCpg != null && !serializedCpg.isEmpty && inverse
+      val doSerialize = serializedCpg != null && !serializedCpg.isEmpty
+      val appliedDiffGraph = DiffGraph.Applier.applyDiff(diffGraph, cpg, withInverse, keyPool)
+      if (doSerialize) {
+        store(serialize(appliedDiffGraph, withInverse), generateOutFileName(prefix, outName, 0), serializedCpg)
+      }
+    } catch {
+      case exc: Exception =>
+        baseLogger.error(s"Enhancement pass ${name} failed", exc)
+        throw exc
+    } finally {
+      try {
+        finish()
+      } finally {
+        // the nested finally is somewhat ugly -- but we promised to clean up with finish(), we want to include finish()
+        // in the reported timings, and we must have our final log message if finish() throws
+        val toc = System.nanoTime()
+        MDC.put("time", s"${(toc - tic) * 1e-6}%.0f")
+        val fracRun = if (tocBuild == -1) 100.0 else (tocBuild - tic) * 100.0 / (toc - tic + 1)
+        baseLogger.info(
+          f"Enhancement $name completed in ${(toc - tic) * 1e-6}%.0f ms (${fracRun}%.0f%% on mutations). ${size}%d changes commited from ${nParts}%d parts.")
+        MDC.remove("time")
+      }
+    }
+  }
+}
+
 trait CpgPassBase {
 
-  private val logger: Logger = LoggerFactory.getLogger(classOf[CpgPass])
+  protected val baseLogger: Logger = LoggerFactory.getLogger(classOf[CpgPass])
 
   def createAndApply(): Unit
 
@@ -123,14 +203,14 @@ trait CpgPassBase {
   }
 
   protected def withStartEndTimesLogged[A](fun: => A): A = {
-    logger.info(s"Start of enhancement: $name")
+    baseLogger.info(s"Start of enhancement: $name")
     val startTime = System.currentTimeMillis
     try {
       fun
     } finally {
       val duration = (System.currentTimeMillis - startTime).millis.toCoarsest
       MDC.put("time", duration.toString())
-      logger.info(s"Enhancement $name completed in $duration")
+      baseLogger.info(s"Enhancement $name completed in $duration")
       MDC.remove("time")
     }
   }
--- a/codepropertygraph/src/main/scala/io/shiftleft/passes/CpgPass.scala
+++ b/codepropertygraph/src/main/scala/io/shiftleft/passes/CpgPass.scala
@@ -84,10 +84,20 @@ abstract class CpgPass(cpg: Cpg, outName: String = "", keyPool: Option[KeyPool] 
 
 }
 
-abstract class NewStylePass[T <: AnyRef](cpg: Cpg, outName: String = "", keyPool: Option[KeyPool] = None)
+abstract class NewStyleSimpleCpgPass(cpg: Cpg, outName: String = "", keyPool: Option[KeyPool] = None)
+    extends NewStyleCpgPass[AnyRef](cpg, outName, keyPool) {
+
+  def run(builder: DiffGraph.Builder): Unit
+
+  final override def generateParts(): Array[_ <: AnyRef] = Array[AnyRef](null)
+
+  final override def runOnPart(builder: DiffGraph.Builder, part: AnyRef): Unit = run(builder)
+}
+
+abstract class NewStyleCpgPass[T <: AnyRef](cpg: Cpg, outName: String = "", keyPool: Option[KeyPool] = None)
     extends CpgPassBase {
   //generate Array of parts that can be processed in parallel
-  def generateParts(): Array[_ <: AnyRef] = Array[AnyRef](null)
+  def generateParts(): Array[_ <: AnyRef]
   //setup large data structures, acquire external resources
   def init(): Unit = {}
   //release large data structures and external resources
@@ -101,10 +111,10 @@ abstract class NewStylePass[T <: AnyRef](cpg: Cpg, outName: String = "", keyPool
                                             inverse: Boolean = false,
                                             prefix: String = ""): Unit = {
     baseLogger.info(s"Start of enhancement: $name")
-    val tic = System.nanoTime()
+    val nanosStart = System.nanoTime()
     var nParts = 0
-    var tocBuild = -1L
-    var size = -1
+    var nanosBuilt = -1L
+    var nDiff = -1
     try {
       init()
       val parts = generateParts()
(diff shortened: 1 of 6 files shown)


mpollmeier (commented):
why do you use the java function/BiConsumer/Supplier API?

bbrehm:
>why do you use the java function/BiConsumer/Supplier API?

That's the java stream API / mutable reduction.

The idea is that `stream.collect` calls the supplier to make a diffGraphBuilder on each thread; the threads start accumulating at evenly selected places. Once a thread finishes its work-slice, it checks whether the next or previous workslice is finished; if so, it can combine the accumulators. Otherwise, it can select another thread, and steal work from it, by obtaining a new accumulator from the supplier, and by by bisecting that thread's remaining work-slice.

The big advantages of this scheme are:
1. The number of diffGraphBuilders that get allocated is a small multiple of the concurenncy level (I think typically 1.0 or ~1.5)
2. Inter-thread / inter-core communication is only required when threads finish their slices -- i.e. a small multiple of the number of threads
3. I don't know the complete openjdk internals on how the stream stuff is implemented. The main communication overhead should be one uncontested atomic per work-item (except if openjdk tries to be even more clever). Especially the number of cache coherence / communication events should be O(1) in the number of work-items.
4. If the supply/accumulate/combine are written correctly, then the result is deterministic and the same irregardless of whether we single-thread or multi-thread.

Parallel associative map-reduce rocks, as a pattern!

bbrehm:
Apart from performance considerations, the previous implementation had a very nasty race condition in the ContainsEdgePass : 

The writer thread runs concurrently with the diffgraph generation. Diffgraph generation needs to walk up the AST; however, intermediate points in the AST can get concurrently modified by addition of CONTAINS edges.

If the new CONTAINS edge doesn't fit, then the adjacentNodesWithEdgeProperties gets regenerated and edgeOffsets updated in a loop. This means that we can get into the situation where the reader thread reads partially updated `edgeOffsets` that don't correspond to the adjacentNodesWithEdgeProperties array.

We could fix this on the overflowdb side by having yet another layer of indirection, where the PackedIntArray wrapper also wraps the adjacentNodesWithEdgeProperties array; and if we need to rewrite / relocate, then we allocate an entire new PackedIntArray and replace it.

That way we could guarantee that graph updates are atomic / that we never read inconsistent data even if another thread writes concurrently / that writing of CONTAINS edges doesn't race against reading of AST edges. (we would still have a data race per java spec, but both possible results would be correct and the same; possibly we'd need a `volatile` somewhere)

However, that would be far too clever -- better avoid the race by design.

mpollmeier (approved):

bbrehm:
Some more thoughts:

-I did read up on the jdk stream implementation... and I'm not happy :( The problem is that this is implemented as divide-and-conquer, and has significant leaf size -- i.e. bad parallelism when the workunits are very uneven. We want a version that pays an atomic for every workunit, to allow fine granularity work-stealing. This does not affect the API, though.
-We also could use scala parallel collection `aggregate` -- that does the same thing as java stream collect.
-To give more reasons for this style of change: In the current parallelCpgPass, all workunits race to enqueue their results in the writer-thread. Even if we fixed read-write races, we get irreproducible order in which diffs are enqueued and afterwards written. The effect of this is partially mitigated by the `KeyPool` construction -- that gives predictable node ids. However, the order is still unpredictable, with effects on the order of iterators (output order of e.g. `cpg.method`) and of neighbors (to take the example of CONTAINS edges: `someMethod._containsOut` has edges to all the ast-nodes inside of the method; the order of this iterator will depend on the outcome of the races).

codepropertygraph/src/main/scala/io/shiftleft/passes/ParallelCpgPass.scala
```diff
@@ -109,12 +111,130 @@ abstract class ParallelCpgPass[T](cpg: Cpg, outName: String = "", keyPools: Opti
...
+    writerThread.start()
+    try {
+      try {
+        var done = false
+        while (!done) {
+          if (completionQueue.size < producerQueueCapacity && partIter.hasNext) {
+            val next = partIter.next()
+            //todo: Verify that we get FIFO scheduling; otherwise, do something about it.
+            //if this e.g. used LIFO with 15 threads, then they work on the last 15 parts, while the main thread blocks on the first;
+            //at some time, the 15 threads finish, one starts on the first part while the remaining 14 idle;
```
bbrehm, line 176: remaining 15 idle

bbrehm (commented):
per zoom, thanks ml86

mpollmeier:
> the previous implementation had a very nasty race condition in the ContainsEdgePass :

> The writer thread runs concurrently with the diffgraph generation. Diffgraph generation needs to walk up the AST; however, intermediate points in the AST can get concurrently modified by addition of CONTAINS edges.

> If the new CONTAINS edge doesn't fit, then the adjacentNodesWithEdgeProperties gets regenerated and edgeOffsets updated in a loop. This means that we can get into the situation where the reader thread reads partially updated edgeOffsets that don't correspond to the adjacentNodesWithEdgeProperties array.

Since we'll talk about this tomorrow, I just looked into this one - and I don't think it's true. 
Background: whenever we grow the array, we create a new one and copy the content via System.arraycpy:
https://github.com/ShiftLeftSecurity/overflowdb/blob/58f0804b68dd44dc0a80660f7ed8d95a6265a43c/core/src/main/java/overflowdb/NodeDb.java#L710

Any previously created reader threads still have the reference to the "old" array and continue iterating that - there's no cross contamination or race when we grow the array afaics. https://github.com/ShiftLeftSecurity/overflowdb/blob/master/core/src/main/java/overflowdb/util/DummyEdgeIterator.java#L22

Maybe I'm missing something, please correct me if I'm wrong. 
Another indicator: if there was a race condition at that place, it's very likely that we would have seen it by now - the current system with ParallelCpgPass and ODB's adjacentNodes array as it is hasn't seen many (any?) changes in the last 1-2 years.

ml86:
The adjacency array is set [here](https://github.com/ShiftLeftSecurity/overflowdb/blob/58f0804b68dd44dc0a80660f7ed8d95a6265a43c/core/src/main/java/overflowdb/NodeDb.java#L653)
and the edgeOffsetArray is set [here](https://github.com/ShiftLeftSecurity/overflowdb/blob/58f0804b68dd44dc0a80660f7ed8d95a6265a43c/core/src/main/java/overflowdb/NodeDb.java#L658).

Since the read operations are not locked at all, a reader can see the new adjacency array with an old edgeOffsetArray.

mpollmeier:
Ok yes that's actually the case then. If a DummyEdgeIterator happens to be instantiated in between these two lines, i.e. after the array was grown but before the edgeOffsets have been updated, then we'd get into the race. 

How about we replace the whole-method (storeAdjacentNode) synchronized block with a mutex (ReentrantLock?) and also lock on that when [creating a DummyEdgeIterator](https://github.com/ShiftLeftSecurity/overflowdb/blob/58f0804b68dd44dc0a80660f7ed8d95a6265a43c/core/src/main/java/overflowdb/NodeDb.java#L575)?

bbrehm:
>How about we replace the whole-method (storeAdjacentNode) synchronized block with a mutex (ReentrantLock?) and also lock on that when creating a DummyEdgeIterator?

It's not just about DummyEdgeIterator, its also about adjacentNodeIterator.

We can either have fast concurrent reads of many threads, or we can permit a writer to run concurrently with readers -- leading to bugs and slowness. 

I can tell you what is better: Just never ever write at the same time as reading, and all races go away. We can then also remove all mutexes.

This doesn't work for database-style APIs, but it works for "batched updates only".

fabsx00 (changes requested):
The potential race condition will be addressed in overflowdb as it is desirable for overflowdb operations to be atomic. The remaining problem with this approach is that it requires all edges to be held in memory for the `ContainsEdgePass`. As there are millions of edges even for a medium sized project such as vlc, that's problematic.

bbrehm:
Documentation NewStyleSimpleCpgPass:
```
/* NewStyleSimpleCpgPass is a possible replacement for CpgPass.
 *
 *  Instead of returning an Iterator[DiffGraph], the `run` fuction gets a DiffGraphBuilder as input, and can attach its
 *  modifications to it (i.e. mutate the builder).
 *
 * CpgPass has somewhat subtle semantics with respect to lazy evaluation order of the returned iterator and graph writes.
 * The subtleties are gone with NewStyleSimpleCpgPass.
 *
 * Note that NewStyleSimpleCpgPass does not support lazy evaluation games. Take care before porting passes that
 * write large amounts of data that risk OOM errors.
 *
 * Initialization and cleanup of external resources or large datastructures can be done in the `init()` and `finish()`
 * methods. This may be better than using the constructor or GC, because e.g. SCPG chains of passes construct
 * passes eagerly, and releases them only when the entire chain has run.
 * */
```
This means that this is inadequate for cases where lazy evaluation / iterator semantics are necessary for correctness or for preventing OOM errors. Afaiu DataFlowPass in the closed-source scanner is the only one that makes use of lazy evaluation for preventing OOM errors; and for problematic projects, we have the newSP to handle that.

I hope that `NewStyleCpgPass` becomes the new "default" for parallelizable passes that don't write excessive amounts of stuff: Simple semantics that are easy to reason about, and low overhead per parallel part.
```
/* NewStyleCpgPass is a possible replacement for CpgPass and ParallelCpgPass.
 *
 * Instead of returning an Iterator, generateParts() returns an Array. This means that the entire collection
 * of parts must live on the heap at the same time; on the other hand, there are no possible issues with iterator invalidation,
 * e.g. when running over all METHOD nodes and deleting some of them.
 *
 * Instead of streaming writes as ParallelCpgPass or LargeChunkPass do, all `runOnPart` invocations read the initial state
 * of the graph. Then all changes (accumulated in the DiffGraphBuilders) are merged into a single change, and applied in one go.
 *
 * In other words, the parallelism follows the fork/join parallel map-reduce (java: collect, scala: aggregate) model.
 * The effect is identical as if one were to sequentially run `runOnParts` on all output elements of `generateParts()`
 * in sequential order, with the same builder.
 *
 * This simplifies semantics and makes it easy to reason about possible races.
 *
 * Note that NewStyleCpgPass never writes intermediate results, so one must consider peak memory consumption when
 * porting from ParallelCpgPass. Consider LargeChunkPass when this is a problem.
 *
 * Initialization and cleanup of external resources or large datastructures can be done in the `init()` and `finish()`
 * methods. This may be better than using the constructor or GC, because e.g. SCPG chains of passes construct
 * passes eagerly, and releases them only when the entire chain has run.
```

And finally, we have `LargeChunkPass` that is a possible replacement for `ParallelCpgPass`: No safety rails for the path author against read-write races on the graph; deterministic behavior in absence of user-induced races (writes are ordered); no subtle pitfalls with respect to lazy evaluation / iterator invalidation; no worse that current ParallelCpgPass with respect to peak memory.
```
/* LargeChunkCpgPass is a possible replacement for ParallelCpgPass and NewStylePass.
 *
 * Instead of returning an Iterator, generateParts() returns an Array. This means that the entire collection
 * of parts must live on the heap at the same time; on the other hand, there are no possible issues with iterator invalidation,
 * e.g. when running over all METHOD nodes and deleting some of them.
 *
 * Changes are applied sequentially, in the same order as the output of `runOnParts`, as opposed to `ParallelCpgPass`,
 * where the ordering of change application is non-deterministic. For this reason, LargeChunkCpgPass only accepts a single KeyPool.
 *
 * However, as opposed to NewStylePass, changes are not buffered and applied in one go; instead, they are applied as the respective
 * `runOnPart` finishes, concurrently with other `runOnPart` invocations.
 *
 * Compared to NewStylePass, this avoids excessive peak memory consumption. On the other hand, `runOnPart` sees the CPG
 * in an intermediate state: No promises are made about which previous changes are already applied; and changes are
 * applied concurrently, such that all reads from the graph are potential race conditions. Furthermore, this variant has
 * higher constant overhead per part than NewStylePass, i.e. is better suited to passes that create few large diffs.
 *
 *
 * Initialization and cleanup of external resources or large datastructures can be done in the `init()` and `finish()`
 * methods. This may be better than using the constructor or GC, because e.g. SCPG chains of passes construct
 * passes eagerly, and releases them only when the entire chain has run.
 * */
```

mpollmeier (approved):

codepropertygraph/src/main/scala/io/shiftleft/passes/ParallelCpgPass.scala
```diff
@@ -109,12 +111,154 @@ abstract class ParallelCpgPass[T](cpg: Cpg, outName: String = "", keyPools: Opti
...
+      } finally {
+        writer.queue.put(None)
+        writerThread.join()
+        finish()
+      }
+    } finally {
+      // the nested finally is somewhat ugly -- but we promised to clean up with finish(), we want to include finish()
+      // in the reported timings, and we must have our final log message if finish() throws
+      val nanosStop = System.nanoTime()
+      MDC.put("time", s"${(nanosStop - nanosStart) * 1e-6}%.0f")
```
hubertp, line 228: Putting context values doesn't seem to make sense here since you do the same calculation inside the log message?

bbrehm: Yeah. I didn't put the MDC here, I just ported it from the ParallelCpgPass because I did not know whether any monitoring scripts rely on it. But you're right, I'll remove it.

codepropertygraph/src/main/scala/io/shiftleft/passes/CpgPass.scala
```diff
@@ -83,9 +84,136 @@ abstract class CpgPass(cpg: Cpg, outName: String = "", keyPool: Option[KeyPool]
...
+ * methods. This may be better than using the constructor or GC, because e.g. SCPG chains of passes construct
+ * passes eagerly, and releases them only when the entire chain has run.
+ * */
+
+abstract class SimpleCpgPass(cpg: Cpg, outName: String = "", keyPool: Option[KeyPool] = None)
+    extends ForkJoinParallelCpgPass[AnyRef](cpg, outName, keyPool) {
+
+  def run(builder: DiffGraph.Builder): Unit
+
+  final override def generateParts(): Array[_ <: AnyRef] = Array[AnyRef](null)
```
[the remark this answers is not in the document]
bbrehm: well, the superclass calls `generateParts()` and then calls `runOnPart` on each array element.

The way I did this, `SimpleCpgPass` doesn't "look into" the `ForkJoinParallelCpgPass` implementation -- it simply is a perfectly valid subclass that uses the public API, and that could be done the same way manually by pass authors who don't want to use the parallel infrastructure. It just takes away the boilerplate.

We certainly could organize this differently -- any suggestions?

codepropertygraph/src/main/scala/io/shiftleft/passes/ParallelCpgPass.scala
```diff
@@ -86,7 +88,7 @@ abstract class ParallelCpgPass[T](cpg: Cpg, outName: String = "", keyPools: Opti
 
     case class DiffGraphAndKeyPool(diffGraph: Option[DiffGraph], keyPool: Option[KeyPool])
 
-    private val logger = LoggerFactory.getLogger(classOf[Writer])
+    private val writeLogger = LoggerFactory.getLogger(classOf[Writer])
```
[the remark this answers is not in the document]
bbrehm: Changed it to use `baseLogger`

codepropertygraph/src/main/scala/io/shiftleft/passes/ParallelCpgPass.scala
bbrehm: removed it

codepropertygraph/src/main/scala/io/shiftleft/passes/ParallelCpgPass.scala
```diff
@@ -109,12 +111,154 @@ abstract class ParallelCpgPass[T](cpg: Cpg, outName: String = "", keyPools: Opti
...
+ * applied concurrently, such that all reads from the graph are potential race conditions. Furthermore, this variant has
+ * higher constant overhead per part than NewStylePass, i.e. is better suited to passes that create few large diffs.
+ *
+ *
+ * Initialization and cleanup of external resources or large datastructures can be done in the `init()` and `finish()`
+ * methods. This may be better than using the constructor or GC, because e.g. SCPG chains of passes construct
+ * passes eagerly, and releases them only when the entire chain has run.
+ * */
+object ConcurrentWriterCpgPass {
+  val writerQueueCapacity = 4
```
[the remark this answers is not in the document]
bbrehm: done

fabsx00 (approved):

PR #1329, opened 2021-08-05
Add the Elvis operator to schema

ursachec:
We're missing this one, and it's found in common patterns when using
languages like Kotlin for example.


fabsx00 (approved):

mpollmeier:
:guitar:

PR #1336, opened 2021-08-06
clean up types and edge api for cfg analysis

bbrehm:
No functional changes. Avoids some casts by choosing more appropriate type bound. Uses the faster API for accessing neighbors.

### DIFF
--- a/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/passes/cfgdominator/CfgDominatorPass.scala
+++ b/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/passes/cfgdominator/CfgDominatorPass.scala
@@ -1,7 +1,8 @@
 package io.shiftleft.semanticcpg.passes.cfgdominator
 
 import io.shiftleft.codepropertygraph.Cpg
-import io.shiftleft.codepropertygraph.generated.{EdgeTypes, nodes}
+import io.shiftleft.codepropertygraph.generated.EdgeTypes
+import io.shiftleft.codepropertygraph.generated.nodes._
 import io.shiftleft.passes.{DiffGraph, ParallelCpgPass}
 import io.shiftleft.semanticcpg.language._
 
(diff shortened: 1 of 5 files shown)


mpollmeier (commented):
I'm ok with the change of types to avoid casts, just one style thing:
We changed our style from `nodes.Xyz` to just `Xzy` to address the node classes a few months ago, can we please keep it that way? There are only two exceptions:
1) extension classes that have the exact same name as the node type, e.g. `dataflowengine.language.extensions.Method` - these cases will swap over when we move to scala 3, the extension type won't need a name any longer then. 
 with the (hopefully) only exception 
2) places I oversaw in the initial swap - i keep on finding more over time

Can you please change that?

mpollmeier:
> apparently I missed the meeting and my chance for voicing objections

yes you did... :hand_over_mouth: 
https://github.com/ShiftLeftSecurity/codepropertygraph/issues/1233

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/passes/cfgdominator/CfgDominatorPass.scala
```diff
@@ -2,10 +2,9 @@ package io.shiftleft.semanticcpg.passes.cfgdominator
 
 import io.shiftleft.codepropertygraph.Cpg
 import io.shiftleft.codepropertygraph.generated.EdgeTypes
-import io.shiftleft.codepropertygraph.generated.nodes.{Method, StoredNode}
+import io.shiftleft.codepropertygraph.generated.nodes._
```
mpollmeier, line 5: in general it's preferrable to avoid wildcard imports, to avoid surprises and improve readability
there are some exceptions, most importantly when importing implicits of course

mpollmeier (approved):

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/passes/cfgdominator/CfgDominatorPass.scala
bbrehm: No. I will not manually import every class used, especially since CI throws errors on unused imports or imports that are not alphabetically ordered or such bullshit.

This is a hill I am willing to die on.

You can have `import io.shiftleft.codepropertygraph.generated.nodes._` everywhere, or you can have ` import io.shiftleft.codepropertygraph.generated.nodes` and `nodes.Method`; choose your poison.

bbrehm: Re #1233 : 

I wholeheartedly agree that renaming the extension classes is a good idea. I don't mind if you write code with explicit imports of specific classes, and then refer to node types with unqualified name.

I strongly reject to making this a mandatory style for all our code, and this was not proposed in https://github.com/ShiftLeftSecurity/codepropertygraph/issues/1233 -- your issue proposed making unqualified names acceptable, not making qualified names unacceptable.

bbrehm: Ok, per slack. not dying on this hill anymore, will fix.

mpollmeier (approved):

PR #1338, opened 2021-08-06
warn when attempting 'reversible irreversible' changes

bbrehm:
There are some changes that we cannot revert. When reversibility is requested by the user but the change cannot be reverted, then warn.

### DIFF
--- a/codepropertygraph/src/main/scala/io/shiftleft/passes/DiffGraph.scala
+++ b/codepropertygraph/src/main/scala/io/shiftleft/passes/DiffGraph.scala
@@ -3,6 +3,7 @@ package io.shiftleft.passes
 import io.shiftleft.codepropertygraph.Cpg
 import io.shiftleft.codepropertygraph.generated.nodes.{AbstractNode, NewNode, StoredNode}
 import io.shiftleft.proto.cpg.Cpg.{DiffGraph => DiffGraphProto}
+import org.slf4j.LoggerFactory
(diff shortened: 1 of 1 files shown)


mpollmeier (commented):
Wait, why are `RemoveEdge|RemoveNode|RemoveNodeProperty` irreversible?

bbrehm:
Because nobody ever implemented the logic to reverse these operations? I think no one requested the feature? AFAIU Denis implemented the serialize / invert logic for diffgraphs, and he left the company more than a year ago.

To be fair, I am somewhat doubtful that the "unapply diffgraph" feature has ever had any users -- I mean, it never worked reliably in the history of this project, and people can just keep a copy of their graph if they forsee needing to revert changes (hard disk is cheap).

I am about to touch / refactor / port-to-java this code, so I wanted to document the gaps beforehand -- and if anyone used the not-implemented codepaths, we hopefully get a bug report and can then consider implementing it (better than just current behavior, i.e. silent failure).

mpollmeier:
I'm pretty sure it's used in the repl/console for undoing overlays. 
Anyway - I fully agree that it's better to warn the user (or even error?) that the implementation is missing, rather than silently fail. 
Maybe just change the wording to 'not yet implemented'? Claiming that these actions are irreversible would confuse me as a user.

bbrehm:
Will be superseded shortly

PR #1345, opened 2021-08-11
clarify that RETURN nodes need to connect to returnee via both AST and ARGUMENT

bbrehm:
Result of a debugging session re frontend/dataflow interactions. Better document that pitfall!


max-leuthaeuser (approved):

bbrehm:
...ok, this is already documented in 
```
    val argument = builder
      .addEdgeType(
        name = "ARGUMENT",
        comment = """Argument edges connect call sites (node type `CALL`) to their arguments
            |(node type `EXPRESSION`) as well as `RETURN` nodes to the expressions
            |that return.
            |""".stripMargin
      )
```

So I just failed at RTFM here, no need to change the docs.

PR #1351, opened 2021-08-16
Add pass chain for KOTLIN

ursachec:


max-leuthaeuser (approved):

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/layers/Scpg.scala
```diff
@@ -140,6 +140,19 @@ class Scpg(optionsUnused: LayerCreatorOptions = null) extends LayerCreator {
...
+          new CfgCreationPass(cpg),
+          new NamespaceCreator(cpg),
+          new FileCreationPass(cpg),
+          new TypeDeclStubCreator(cpg),
+          new ContainsEdgePass(cpg),
+          new Linker(cpg),
+          new MethodDecoratorPass(cpg),
+          new StaticCallLinker(cpg),
+          new CfgDominatorPass(cpg),
+          new CdgPass(cpg),
```
ml86, line 154: Can you please use reorder the passes to be as close as possible to the other pass chain definitions?
Otherwise this just creats confusion.

ursachec: yes, good point, reordered them here https://github.com/ShiftLeftSecurity/codepropertygraph/pull/1352

PR #1363, opened 2021-08-19
Add HeuristicsCallLinker

ursachec:


semanticcpg/src/main/scala/io/shiftleft/semanticcpg/passes/linking/calllinker/HeuristicsCallLinker.scala
```diff
@@ -0,0 +1,102 @@
...
+    if (resolvedMethodOption.isDefined) {
+      if (resolvedMethodOption.get.size == 1) {
+        dstGraph.addEdgeInOriginal(call, resolvedMethodOption.get(0), EdgeTypes.CALL)
+      } else {
+        val withNumParams =
+          resolvedMethodOption.get
+            .map(_.asInstanceOf[Method])
+            .map { method =>
+              method.parameter.size -> method
+            }
```
ml86, line 78: Dont do this for each call site. Do it only once per method during creation of `methodNameToNodes` and store a sequence of tuples instead of just a sequence of methods.

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/passes/linking/calllinker/HeuristicsCallLinker.scala
```diff
@@ -0,0 +1,102 @@
...
+          withNumParams.filter { withNum =>
+            withNum._1 == call.argument.size
+          }
+        if (matchingNumOfParams.size > 0) {
+          matchingNumOfParams.foreach { entry =>
+            dstGraph.addEdgeInOriginal(call, entry._2, EdgeTypes.CALL)
+          }
+        } else {
+          // TODO: use logger
+          println(
```
ml86, line 90: Use logger.

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/passes/linking/calllinker/HeuristicsCallLinker.scala
```diff
@@ -0,0 +1,102 @@
...
+            s"INFO: Unable to link static CALL with METHOD_FULL_NAME ${call.methodFullName}, NAME ${call.name}, " +
+              s"SIGNATURE ${call.signature}, CODE ${call.code}")
+        }
+      }
+    } else {
+      // TODO: use logger
```
ml86, line 96: Use logger.

ml86 (changes requested):
Move hole pass to kotlin specific name space. With the full name assumptions it is not generic. If we make it generic at some point we can move it again.

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/passes/linking/calllinker/HeuristicsCallLinker.scala
ursachec: yes, much better idea, thanks!

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/passes/languagespecific/kotlin/HeuristicsCallLinker.scala
```diff
@@ -0,0 +1,92 @@
...
+    } else {
+      logger.info(
+        s"Unable to link static CALL with METHOD_FULL_NAME ${call.methodFullName}, NAME ${call.name}, " +
+          s"SIGNATURE ${call.signature}, CODE ${call.code}")
+    }
+  }
+}
+
+object HeuristicsCallLinker {
+  private val logger: Logger = LoggerFactory.getLogger(classOf[HeuristicsCallLinker])
```
ml86, line 91: Use `getClass` instead of `classOf[...]`.

ml86 (approved):

PR #1364, opened 2021-08-19
invoke fuzzyc2cpg via shell command rather than in running jvm - to avoid memory leak

mpollmeier:
I noticed that when analysing medium sized codebases (e.g.
`linux-4.1.16/drivers/staging` which is 13M), after running
`importCode`, there was still a large `io.shiftleft.fuzzyc2cpg.FunctionParser`
instance, allocating 330M on heap. I'll attach a screenshot from
the profiler which also shows the path to GC.
The class is generated via antlr and it wasn't immediately clear how to
avoid this memory leak.

Since the cpg was closed and reopened before as well, this change shouldn't
have much of a time/efficiency effect.

yourkit-joern-fuzzyc-heap


ml86 (approved):

fabsx00:
Before bringing this into joern/ocular, please also verify that this works when they have been installed, e.g., via `joern-install.sh`. We had problems with this in the past which is why we decided to call this frontend via the JVM instead.

PR #1365, opened 2021-08-19
Fullname and TypeDecl fixes

max-leuthaeuser:
The where tons of defect fullnames and typedecl names/fullnames coming from internal CDT helper classes like ASTTypeUtils.
Had to fix them manually.


fabsx00 (approved):

max-leuthaeuser:
The where tons of defect fullnames and typedecl names/fullnames coming from internal CDT helper classes like ASTTypeUtils.
Had to fix them manually.

PR #1366, opened 2021-08-19
Do not parse header files explicitly

max-leuthaeuser:
This is done by the CDT preprocessor already which mimics the actual gcc behavior (include directive).

This needs to be discussed. We won't have findings in header files directly anymore.
Instead they would appear in the corresponding *.c or *.cpp file (wherever the header is included).


johannescoetzee:
Something to keep in mind for this change is that, when using `joern-scan`, findings in the header file will be reported as being defined in the corresponding c file instead. There's a more detailed example of that in the issue linked above, but effectively it would be reporting something like
```
Result: 1.0 : Multiple returns: test_inline.c:7:vulnerability_in_inline_func
```
instead of
```Result: 1.0 : Multiple returns: test_inline.h:7:vulnerability_in_inline_func
```

Consequently, the file and line number reported for the finding may not be where the issue occurs at all. We'd need some way of identifying whether a finding occurred in a C file or the corresponding header file for accurate reporting.

max-leuthaeuser:
Yes, you are right. Its not that easy here. Lets close this one for now.

Issue #1361, opened 2021-08-18
Warnings and crash when using the new C frontend with joern-scan on VLC


johannescoetzee:
When running joern-scan on the VLC repo using the new frontend, the following warnings were shown in the log file. These warnings were not emitted when running joern-scan using the fuzzy frontend.

```
2021-08-18 09:00:07.711 WARN CustomFileContentProvider: Include file not found: '/home/johannes/code/test_repos/vlc/compat/test/config.h'
```
```
2021-08-18 09:00:08.038 WARN AstCreator: Node 'CASTSimpleDeclaration' not handled yet!
  Code: 'LIBVLC_API'
  File: '/home/johannes/code/test_repos/vlc/include/vlc/libvlc_media.h'
  Line: 701
  ```
  
  ```
  Adding default overlays to base CPG
2021-08-18 09:00:54.328 WARN Linker: Invalid AST_PARENT_TYPE=Optional[<empty>]; astChild LABEL=METHOD; astChild FULL_NAME=Optional[VLCVideo.VLCVideo.anonymous_lambda_0]
2021-08-18 09:00:54.328 WARN Linker: Could not create edge. Source lookup failed. edgeType=AST, srcNodeType=<empty>, srcFullName=<empty>, dstNodeType=METHOD, dstNodeId=1005246
```

```
2021-08-18 09:00:55.145 WARN Linker: Could not create edge. Destination lookup failed. edgeType=EVAL_TYPE, srcNodeType=METHOD_REF, srcNodeId=1005366, dstNodeType=TYPE, dstFullName=VLCVideo.VLCVideo.anonymous_lambda_0
```

```
2021-08-18 09:00:59.574 WARN MemberAccessLinker: Error while obtaining IDENTIFIER associated to member access at io.shiftleft.codepropertygraph.generated.nodes.Call[label=CALL; id=1005157] Reason: member/field access-style operators need two arguments
```

```
2021-08-18 09:01:24.762 WARN AccessPathUsage$: Missing first argument on call io.shiftleft.codepropertygraph.generated.nodes.Call[label=CALL; id=1010985].
```

The run then ended with a crash due to the following error:

```
Exception in thread "main" java.lang.AssertionError: script errored: 
	at io.shiftleft.console.BridgeBase.runScript(BridgeBase.scala:356)
	at io.shiftleft.console.BridgeBase.$anonfun$runPlugin$4(BridgeBase.scala:242)
[...]

PR #1367, opened 2021-08-19
Back to static dispatch

max-leuthaeuser:
This allows to run CPGs through ocular completely now.
Dynamic dispatch is TODO.

Also we are not allowed to remove method stubs as they might be required for typedecl bindings / refs.


max-leuthaeuser:
E.g., this query runs now smoothly: https://github.com/ShiftLeftSecurity/codepropertygraph/issues/1361#issuecomment-901111179

johannescoetzee (approved):
LGTM, but this comes without any knowledge about Ocular on my part. Since we don't use c2cpg for production yet, I see little risk.

fabsx00 (approved):
The `StubRemovalPass` is a hack that deals with intrinsics of fuzzyc2cpg, so, this seems fine.

PR #1376, opened 2021-08-24
Make frontend config buffer immutable

johannescoetzee:
# Notes
* Make cmdLineParams immutable in FrontendConfig
* I'm keeping the `withAdditionalArgs` method as it's not quite a straight-forward copy (since we want to append the additional args to the existing ones). I did modify the implementation to use copy to make this more explicit, however.
# Testing
`sbt clean test`


ursachec (approved):

hubertp (approved):
Thanks, lgtm

PR #1381, opened 2021-08-27
Support for more C/C++ AST elements

max-leuthaeuser:
Added support for:

- friend declarations
- pack expansion
- varargs
- namespace aliases in method bodies
- ASM declarations
- Structured Binding Declarations
- Using Directives in method bodies

### DIFF
--- a/c2cpg/src/main/scala/io/shiftleft/c2cpg/astcreation/AstCreatorHelper.scala
+++ b/c2cpg/src/main/scala/io/shiftleft/c2cpg/astcreation/AstCreatorHelper.scala
@@ -1,5 +1,6 @@
 package io.shiftleft.c2cpg.astcreation
 
+import io.shiftleft.codepropertygraph.generated.DispatchTypes
 import io.shiftleft.x2cpg.Ast
 import org.eclipse.cdt.core.dom.ast._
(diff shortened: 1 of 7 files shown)


c2cpg/src/main/scala/io/shiftleft/c2cpg/astcreation/AstCreatorHelper.scala
```diff
@@ -235,15 +236,29 @@ trait AstCreatorHelper {
     }
   }
 
+  private def astforDecltypeSpecifier(decl: ICPPASTDecltypeSpecifier, order: Int): Ast = {
+    val op = "operators.<typeOf>"
```
fabsx00, line 240: Is this operator defined in the schema, and if so, can we reference as constant?

fabsx00 (approved):

c2cpg/src/main/scala/io/shiftleft/c2cpg/astcreation/AstCreatorHelper.scala
max-leuthaeuser: Sadly, its not. We might want to add it as its quite common.

PR #1383, opened 2021-09-01
Add the cpg generators dir to InstallConfig to not require front-end scripts in project root

johannescoetzee:
# Notes
This PR adds an optional `cpgGeneratorsDir` to `InstallConfig` and uses that field when determining the front-end executable path for `c2cpg`, `fuzzyc2cpg`, `ghidra2cpg`, and `js2cpg` (so effectively the front-ends available through Joern). This can easily be extended to most of the rest of the front-ends, but some thought is needed for Go and C#, since they have Windows options as well.

# Testing
`sbt clean test`, along with manual verification that these scripts no longer need to appear in the Joern project root  for it to function (follow-up PR coming)

### DIFF
--- a/console/src/main/scala/io/shiftleft/console/cpgcreation/CCpgGenerator.scala
+++ b/console/src/main/scala/io/shiftleft/console/cpgcreation/CCpgGenerator.scala
@@ -1,14 +1,20 @@
 package io.shiftleft.console.cpgcreation
 
-import io.shiftleft.console.FrontendConfig
+import io.shiftleft.console.{FrontendConfig, InstallConfig}
 
-import java.nio.file.Path
+import java.nio.file.{Path, Paths}
 
 /**
   * Fuzzy C/C++ language frontend. Translates C/C++ source files
   * into code property graphs via fuzzy parsing.
   * */
-case class CCpgGenerator(config: FrontendConfig, rootPath: Path) extends CpgGenerator {
+case class CCpgGenerator(config: FrontendConfig, installConfig: InstallConfig) extends CpgGenerator {
+
+  val binaryPath: Path = {
+    val rootPath = installConfig.rootPath.path
+    val relativeBinPath = Paths.get(installConfig.cpgGeneratorsDir, "c2cpg.sh")
+    rootPath.resolve(relativeBinPath)
+  }
 
   /**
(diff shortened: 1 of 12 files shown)


console/src/main/scala/io/shiftleft/console/cpgcreation/CCpgGenerator.scala
```diff
@@ -1,14 +1,20 @@
...
+import java.nio.file.{Path, Paths}
 
 /**
   * Fuzzy C/C++ language frontend. Translates C/C++ source files
   * into code property graphs via fuzzy parsing.
   * */
-case class CCpgGenerator(config: FrontendConfig, rootPath: Path) extends CpgGenerator {
+case class CCpgGenerator(config: FrontendConfig, installConfig: InstallConfig) extends CpgGenerator {
+
+  val binaryPath: Path = {
```
ursachec, line 13: since the code for `binaryPath` is repeating, maybe it would make sense to put it inside _CpgGenerator_ and have the script name instead be specific for the various subclasses?

johannescoetzee: Funny you should mention that. I originally did exactly that. The issue I ran into there was dealing with the front-ends with Windows options in a similar way, but 2 options there is to have a separate `winPath` for all the front-ends (which would be empty for most of them), or to hack it in for the 2 others. Instead of deciding on that now, I opted to go for the minimal change in this case, with ideas of cleaning everything up properly later.

An option I didn't consider is to define the `binaryPath` in `CpgGenerator`, but to only use that for the *nice* frontends (which is already better than what I have here.

ursachec: alrighty, if you want to deal with Windows at a later point, :+1: from my side.
the one thing I'd still adjust though is the name of the val, maybe `invocationScriptPath` is a better description?

johannescoetzee: That's a much better name!

johannescoetzee:
No longer needed due to https://github.com/ShiftLeftSecurity/codepropertygraph/pull/1387

PR #1395, opened 2021-09-09
Better `code` fields for expanded macros

fabsx00:
For code such as the following:
```
#define A_MACRO(x) (x = 10)
       int foo() {
        int y;
        A_MACRO(y);
        return 10 * y;
       }
```
the macro expansion `A_MACRO(y)` was represented in the CPG as an assignment node with two children, all of which had a code field of `A_MACRO(y)`, which is a bit confusing. The reason is that the `code` field refers to the code _without_ considering the preprocessor as obtained by `getRawSignature` while the nodes represent the AST as obtained by expanding the macro.

With this PR, we now no longer call `getRawSignature` directly but instead call our own utility method `AstCreator.nodeSignature`. This method checks whether the node refers to a macro expansion. If that is not the case, we simply call `getRawSignature`. If it is, then we call a utility method from Eclipse CDT that translates AST nodes into strings.

The result is that, whether we use a macro or not, the AST looks the same, that is, the AST is the same as we would obtain for
```
       int foo() {
        int y;
        y = 10;
        return 10 * y;
       }
```

The CDT utility method is unfortunately in a class that is marked as deprecated, however, this is because the CDT team would like to discourage its use but at the same time does not plan to remove this code. I would suggest we make use of it for now and see how that goes. If we run into problems, we will need to implement similar functionality ourselves.

A problem with this fix is that we now no longer see that the macro was used at all, and that's unfortunate because macros such as `GET32` or `n2s` make great sources. I will deal with this problem in a follow-up PR.


c2cpg/src/main/scala/io/shiftleft/c2cpg/astcreation/AstCreator.scala
```diff
@@ -106,3 +107,17 @@ class AstCreator(val filename: String, val global: Global, val config: C2Cpg.Con
   }
 
 }
+
+object AstCreator {
+
+  @nowarn
```
mpollmeier, line 113:
```suggestion
  @nowarn
    /** The CDT utility method is unfortunately in a class that is marked as deprecated, however, 
    * this is because the CDT team would like to discourage its use but at the same time does 
    * not plan to remove this code.
    */
```

mpollmeier (approved):

max-leuthaeuser (approved):

johannescoetzee (approved):

PR #1406, opened 2021-09-15
Improved Macro Handling for new C/C++ frontend

fabsx00:
At construction time of the AST, we now determine sub trees that are created via macro expansion and introduce a CALL node that represents the reference to the macro. Since this isn't really a call but an inlining of code, I introduced a new dispatch type called `INLINED` for these calls. The expansion sub tree is included as a child of this call node. Moreover, for each macro argument, we include an Identifier node where the `CODE` field contains the argument as a string. In a second iteration, we would like to instead include the arguments as sub trees rather than strings, however, this is currently blocked on a missing feature in overflowdb (cloning of NewNodes).

So, why a CALL for macro expansions? Because a macro expansion looks like a call to the person reading the code and I can already see them scratching their head when no calls come back for the name they just saw in the code. At the same time, leaving the macro unexpanded makes our analysis much more precise than in fuzzyc times.

Changes in detail:
- Introduced MacroArgumentExtractor, which extracts arguments of macros via CDT voodoo.
- Concentrated all handling of macros done in the AST construction stage into `MacroHandler`, which is now a mixin trait just like `AstCreatorHelper`.
- The CFG creator now picks up calls with dispatch type `INLINED` and represents them as calls to the macro followed by optional evaluation of the inlined code.
- I noticed that for simple `f(x)` calls, we included a receiver of `f`. Two things that come to mind: we end up with many rather redundant identifier nodes that simply contain `f` although `f` is already in the `NAME` field of the call. Given that the CPG mainly consists of call sites, this seems significant. The second reason I think it makes sense to remove these nodes is that the data flow calculation will then track `f`, which slows it down considerably.

Some known/intended limitations:
- If there is a macro used in a macro, we don't see that in the AST of the callee, we simply see the entire expansion. This seems alright though because the whole reason for making available the call to the macro is that this call is something the auditor looking at the code sees and if they only have the expansion they may be confused about that. The fact that the expansion is actually created via multiple expansions seems less relevant.
- AST rewriting is pretty hard without the copy method for `NewNode` classes, so, currently, the order/argument_index field of expanded nodes inserted below artificial call nodes is possibly wrong. I added a TODO.

### DIFF
--- a/c2cpg/src/main/scala/io/shiftleft/c2cpg/astcreation/MacroHandler.scala
+++ b/c2cpg/src/main/scala/io/shiftleft/c2cpg/astcreation/MacroHandler.scala
@@ -93,7 +93,7 @@ trait MacroHandler {
 
     val argAsts = arguments.zipWithIndex.map {
       case (arg, i) =>
-        Ast(NewIdentifier(code = arg, order = i + 1))
+        Ast(NewIdentifier(name = arg, code = arg, order = i + 1, argumentIndex = i + 1))
     }
 
     Ast(callNode)
--- a/c2cpg/src/main/scala/io/shiftleft/c2cpg/astcreation/MacroHandler.scala
+++ b/c2cpg/src/main/scala/io/shiftleft/c2cpg/astcreation/MacroHandler.scala
@@ -3,7 +3,13 @@ package io.shiftleft.c2cpg.astcreation
 import io.shiftleft.codepropertygraph.generated.DispatchTypes
 import io.shiftleft.codepropertygraph.generated.nodes.NewCall
 import io.shiftleft.x2cpg.Ast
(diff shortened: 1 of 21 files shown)


c2cpg/src/main/scala/io/shiftleft/c2cpg/astcreation/MacroHandler.scala
```diff
@@ -0,0 +1,141 @@
...
+                           node,
+                           mac,
+                           args)
+    }
+    if (macroCallAst.isDefined) {
+      macroCallAst.get.withChild(ast)
+    } else {
+      ast
+    }
+  }
```
mpollmeier, line 34:
```suggestion
  def asChildOfMacroCall(node: IASTNode, ast: Ast): Ast =
    extractMatchingMacro(node).map { case (mac, args) =>
      createMacroCallAst(node, mac, args)
    } match {
      case Some(macroCallAst) => macroCallAst.withChild(ast)
      case _ => ast
    }
```

mpollmeier (approved):

PR #1410, opened 2021-09-17
also use current working dir in search for installation root - next try

mpollmeier:
new attempt: first search in lib dir, then cwd; also add docs

determining the root path of the joern/ocular installation is rather complex unfortunately,
because we support a variety of use cases:
- running the installed distribution from the install dir
- running the installed distribution anywhere else on the system
- running a locally staged ocular/joern build (via `sbt stage` and then
  either `./joern` or `cd joern-cli/target/universal/stage; ./joern`)
- running a unit/integration test (note: the jars would be in the local cache, e.g. in ~/.coursier/cache)

### DIFF
--- a/console/src/main/scala/io/shiftleft/console/ConsoleConfig.scala
+++ b/console/src/main/scala/io/shiftleft/console/ConsoleConfig.scala
@@ -10,18 +10,27 @@ import scala.collection.mutable
   * @param environment A map of system environment variables.
   * */
 class InstallConfig(environment: Map[String, String] = sys.env) {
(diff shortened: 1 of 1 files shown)


mpollmeier:
and yes, we should really add an integration test like 'testDistro.sh' to joern - it's on my list :)

fabsx00 (approved):
Approving with fear in my eyes ;)

ursachec (approved):

mpollmeier:
🔥 👁️

PR #1414, opened 2021-09-17
Search for header files in project directory if not found

fabsx00:
If a header file can't be found, check if it is anywhere in the project directory. If there are multiple files with that name, choose the one with the lowest edit distance to the path of the source file (not 100% true, it's `/path/to/source/file/header.h`, but that should be the same qualitatively).

### DIFF
--- a/c2cpg/src/main/scala/io/shiftleft/c2cpg/parser/HeaderFileFinder.scala
+++ b/c2cpg/src/main/scala/io/shiftleft/c2cpg/parser/HeaderFileFinder.scala
@@ -11,7 +11,8 @@ class HeaderFileFinder(roots: List[String]) {
   private val headerExtensions = List(".h", ".hpp", ".hh")
   private val nameToPathMap: Map[String, List[Path]] = {
 
-    SourceFiles.determine(roots.toSet, headerExtensions.toSet)
+    SourceFiles
+      .determine(roots.toSet, headerExtensions.toSet)
       .map { p =>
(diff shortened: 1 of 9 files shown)


max-leuthaeuser (approved):

c2cpg/src/main/scala/io/shiftleft/c2cpg/parser/HeaderFileFinder.scala
```diff
@@ -0,0 +1,39 @@
...
+
+import java.nio.file.{FileSystems, Path}
+
+class HeaderFileFinder(roots: List[String]) {
+
+  private val headerExtensions = List(".h", ".hpp", ".hh")
+  private val nameToPathMap: Map[String, List[Path]] = {
+    roots
+      .flatMap { root =>
+        File(root).listRecursively.filter(f => f.extension.isDefined && headerExtensions.contains(f.extension.get))
```
max-leuthaeuser, line 14: It might be quite slow doing this.
We could also reuse this: https://github.com/ShiftLeftSecurity/codepropertygraph/blob/master/c2cpg/src/main/scala/io/shiftleft/c2cpg/C2Cpg.scala#L42

fabsx00: You're right, I've written a better version before ;) - Reusing `SourceFiles.determine`

PR #1426, opened 2021-09-27
Resolve FileSystems.newFileSystem overload ambiguity on Java 13+

DavidBakerEffendi:
The compiler complains about `FileSystems.newFileSystem` overload ambiguity associated with extra `newFileSystem` overloaded methods.  [Java 11 `newFileSystem`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/nio/file/FileSystems.html) has 3 overloaded methods whereas [Java 13+ `newFileSystem`](https://docs.oracle.com/en/java/javase/13/docs/api/java.base/java/nio/file/FileSystems.html) has 6.

The kind of exception thrown on the `ScriptManagerTest` also varies and was adjusted.

This PR aims to resolve that by giving a type to the `null` argument.

### DIFF
--- a/codepropertygraph/src/main/scala/io/shiftleft/codepropertygraph/cpgloading/ZipArchive.scala
+++ b/codepropertygraph/src/main/scala/io/shiftleft/codepropertygraph/cpgloading/ZipArchive.scala
@@ -8,7 +8,8 @@ import scala.collection.mutable.ArrayBuffer
 import scala.jdk.CollectionConverters._
 
 class ZipArchive(inputFile: String) extends Closeable {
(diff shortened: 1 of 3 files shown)


mpollmeier (approved):
thanks, I also tested on jdk16 and 17
one test is red on those though: https://github.com/ShiftLeftSecurity/codepropertygraph/blob/master/codepropertygraph/src/test/scala/io/shiftleft/codepropertygraph/cpgloading/CpgLoaderTests.scala#L41 needs the same jvm version switch as you added for `ScriptManagerTest`

To simplify matters, I'd suggest to change both places to simply `intercept[Exception]` or similar.

PR #1438, opened 2021-10-05
Introduce ASTs for arguments of macro expansions

fabsx00:
In this PR, we make use of overflowdb's new node copying mechanism to create copies of ASTs for the arguments of macro expansions. For example, consider the following expansion:

```
#define MY_MACRO(x) (x) = 1

MY_MACRO(*foo);
```

For this expansion, we now create a call to `MY_MACRO` with one argument that is an AST consisting of a call to `Operator.Indirection` with the argument `foo`.


johannescoetzee (approved):

fabsx00:
The `METHOD` nodes are still missing and require a follow-up PR which I am working on right now.

PR #1476, opened 2021-10-21
Remove `MemberAccessLinker`

fabsx00:
The `MemberAccessLinker` creates REF edges from calls to members, and these edges seem to not be used in the OSS world. I am therefore moving the `MemberAccessLinker` into codescience and removing invocations from `Scpg`.


mal-tee:
Hi, where can I find `codescience`? I am trying to upgrade a project from "1.3.122" and have to fix some imports, e.g. moved passes.

fabsx00:
@mal-tee that's a commercial repo, but what this means is that the `MemberAccessLinker` is no longer required in Joern.

mal-tee:
Thank you for the info. Pinging @simkoc

PR #1502, opened 2021-11-03
Removed depedency to small library.

ml86:
Remove dependency to a not very popular library and used scala standard
library classes instead.

Also updated scalatest and json4s dependency versions.

### DIFF
--- a/project/Versions.scala
+++ b/project/Versions.scala
@@ -1,7 +1,7 @@
 /* reads version declarations from /build.sbt so that we can declare them in one place */
 object Versions {
   val overflowdb = parseVersion("overflowdbVersion")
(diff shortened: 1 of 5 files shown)


bbrehm (approved):

mpollmeier:
collection-contrib is used in codescience, and upgrading cpg leads to a few compilation issues due to that. 
Can you please do that upgrade and decide if we want to remove the usage of MultiDict etc. in codescience as well, or alternatively add it back as a dependency to e.g. semanticcpg-ext?

ml86:
Yep. For now i will just add it as a dependency there which is how it should have been in the first place.

PR #1504, opened 2021-11-04
Write INFO and lower logs to file instead of console when running tests

johannescoetzee:
The CPG logs are currently written to the console when running tests, which is quite noisy and makes it difficult to identify test failures from `sbt test` output. With this change, only `ERROR` level logs are output to the console, while the rest of the logs are written to a file instead.

Another option I considered was dropping these logs to `DEBUG` and only printing `INFO` or higher logs. This is a perfectly fine option too, although having these logs available is nice as a cheap means of profiling.

### DIFF
--- a/resources/log4j2-test.xml
+++ b/resources/log4j2-test.xml
@@ -4,6 +4,9 @@
         <Console name="Console" target="SYSTEM_OUT">
             <PatternLayout pattern="[%-5p] %m%n" />
         </Console>
+        <File name="FileAppender" fileName="${sys:user.dir}/logs/cpg.log">
+            <PatternLayout pattern="[%-5p] %m%n" />
+        </File>
     </Appenders>
     <Loggers>
         <Logger name="org.reflections8.Reflections" level="error"/>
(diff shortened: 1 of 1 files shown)


fabsx00 (approved):

resources/log4j2-test.xml
```diff
@@ -4,6 +4,9 @@
         <Console name="Console" target="SYSTEM_OUT">
             <PatternLayout pattern="[%-5p] %m%n" />
         </Console>
+        <File name="FileAppender" fileName="${sys:user.dir}/logs/cpg.log">
```
mpollmeier, line 7: does that write it to the current working directory, or the user's home dir?

johannescoetzee: It writes to the current working directory. In an ideal case, I would've chosen the cpg/joern/ocular root directory as the target, but I'm not sure how that could be done. It may be possible if you create the appender in code instead of in the config file, but that feels a bit ugly.

johannescoetzee: I think writing to the current working directory is fine when running tests (since those would most likely be run from the project directory in any case), but wouldn't work for general Joern/Ocular logs, for example, since those could be run from anywhere.

mpollmeier: CWD is great, let's keep it simple :)

PR #1505, opened 2021-11-04
Update scala version and scalameta dependency version.

ml86:


mpollmeier (commented):
Hmm, it might be a good idea to wait for 2.13.8 which shouldn't be long...
https://twitter.com/RomanowskiKr/status/1455477298107453447

Also note the comment in build.sbt:10 about issues with ammonite. 
This doesn't apply to this repo any more though, because console moved to joern. 
So we can drop that comment here.

ml86:
Yes, in that case we should wait.

mpollmeier:
apparently this is fixed now in the latest bloop release, so we can go ahead with this
https://twitter.com/RomanowskiKr/status/1455624973121466372

mpollmeier (approved):

PR #1506, opened 2021-11-04
Remove preprocessor for fuzzyc

fabsx00:
Removing the preprocessor experiment we did for fuzzyc as it is barely used, complicates matters dramatically and we now have the newc frontend, which does all of this seamlessly.


mpollmeier (approved):
🎉

PR #1510, opened 2021-11-05
Cleanup node starter steps

fabsx00:
* Remove starter steps that have been deprecated for over a year
* Sort steps by name
* Be consistent about using `name` as opposed to `fullName` in starter steps that accept a name

This may require downstream adaptions. Checking.

PR #1511, opened 2021-11-08
Add control structure type `throw`

johannescoetzee:
The current approach for handling throws is to represent them as a call to a throw operator. The behaviour of a `throw statement is closer to a control structure (specifically `break` and `continue`), however, so it makes more sense to represent this as a CS instead of a call.

### DIFF
--- a/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/NodeTypeStarters.scala
+++ b/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/NodeTypeStarters.scala
@@ -75,6 +75,10 @@ class NodeTypeStarters(cpg: Cpg) {
   def continue: Traversal[ControlStructure] =
     controlStructure.isContinue
 
+  @Doc("All throws (`ControlStructure` nodes)")
+  def throws: Traversal[ControlStructure] =
+    controlStructure.isThrow
+
   /**
     Traverse to all source files
     */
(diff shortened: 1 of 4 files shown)


semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/NodeTypeStarters.scala
```diff
@@ -75,6 +75,10 @@ class NodeTypeStarters(cpg: Cpg) {
   def continue: Traversal[ControlStructure] =
     controlStructure.isContinue
 
+  @Doc("All throws (`ControlStructure` nodes)")
+  def throws: Traversal[ControlStructure] =
```
johannescoetzee, line 79: It would've been better to name this `throw` for consistency with other starters, but this doesn't work since `throw` is a reserved word.

johannescoetzee:
This conflicts with https://github.com/ShiftLeftSecurity/codepropertygraph/pull/1510 (which should rather be merged first)

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/NodeTypeStarters.scala
mpollmeier: makes sense. you could add an escaped alias which, but that's not everyones favorite, understandably
```
def `throw`: Traversa[ControlStructure] = throws
```

mpollmeier (approved):

fabsx00 (approved):

PR #1523, opened 2021-11-11
New traversal proposal.

ml86:


fabsx00:
From your description, the advantages clearly outweight the disadvantages. The one thing I ask myself when seeing the example code is whether we can find any ways of making the complexity less visible to the language writer. @mpollmeier what are your thoughts?

mpollmeier:
Just discussed this in detail with markus, and he's looking into a few additional points (e.g. support for `PathAwareTraversal` and `.repeat()(_.until)` with type application). We'll let it sink in over the weekend. 

Additional benefit that's not in the description yet: the end user (not the language designer) doesn't need to learn about `Traversal` - she can just continue to use `Iterator` or whatever collection they have at hand. 

Additional downside: when the language designer adds a non-trivial step (like in `LocalTraversalNew.referencingIdentifiers`), she only has the methods that we defined in `AnyTraversal` at hand. Currently, with `Traversal`, we have all collection methods, including things like zipWithIndex, partition etc. We'd need to make sure we cover enough ground here. 

We also discussed and agreed that it makes sense to bring this in _before_ upgrading to Scala 3, mostly because changing the Scala version _and_ traversal impl wold mean a hard cut for all existing Scala 2 users. 

It'll be crucial to document this really really well, especially those additional generic terms like InputTraversal(IT) / FlatMapTraversal(IT) and TravNOps. 

Overall I'm positive for this PR. The complexity certainly goes up, but the benefits seem to outweigh the downsides.

bbrehm (changes requested):
Stale, please re-request review once you're ready to continue.

fabsx00 (commented):
Stale

PR #1527, opened 2021-11-15
Logging fixes, pass through execution context.

Ferada:
Essentially a draft and prerequisite for setting up the slf4j MDC for inheriting key-value pairs for all spawned tasks.

### DIFF
--- a/codepropertygraph/src/main/scala/io/shiftleft/passes/CpgPass.scala
+++ b/codepropertygraph/src/main/scala/io/shiftleft/passes/CpgPass.scala
@@ -8,6 +8,7 @@ import org.slf4j.{Logger, LoggerFactory, MDC}
 
 import java.lang.{Long => JLong}
 import java.util.function.{BiConsumer, Supplier}
+import scala.annotation.nowarn
 import scala.concurrent.ExecutionContext
 import scala.concurrent.duration.DurationLong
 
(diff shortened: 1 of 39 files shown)


codepropertygraph/src/main/scala/io/shiftleft/passes/ParallelCpgPass.scala
```diff
@@ -220,15 +217,15 @@ abstract class ConcurrentWriterCpgPass[T <: AnyRef](cpg: Cpg, outName: String =
       } finally {
         try {
           writer.queue.put(None)
-          writerThread.join()
+          Await.ready(task, Duration.Inf)
```
hubertp, line 220: We should probably have some reasonable timeouts but that's beyond the scope of this change since previously essentially it was doing the same thing

hubertp (approved):

codepropertygraph/src/main/scala/io/shiftleft/passes/ParallelCpgPass.scala
Ferada: Yes and agreed.

codepropertygraph/src/main/scala/io/shiftleft/passes/CpgPass.scala
```diff
@@ -34,23 +36,28 @@ abstract class CpgPass(cpg: Cpg, outName: String = "", keyPool: Option[KeyPool]
   /**
     * Main method of pass - to be implemented by child class
     * */
-  def run(): Iterator[DiffGraph]
+  def run(): Iterator[DiffGraph] = ???
```
Ferada, line 39: Not super happy with this. On the other hand I'd also like to avoid touching hundreds of files for a naming change. Any alternatives that force implementation of either `run` or `runWithExecutionContext`?

Ferada: @hubertp

hubertp: Nope no other way if you don't want to touch those files

hubertp (approved):

Ferada:
> Nope no other way if you don't want to touch those files

Depends, I attempted to do it and then quickly realised there's no automatic way, it's tedious, will annoy readers and doesn't actually get us much except this one problem, which we'd hopefully rule out as a problem by testing instead.

PR #1528, opened 2021-11-17
CpgPass with one method required to extend

note:
Just a draft, in response to https://github.com/ShiftLeftSecurity/codepropertygraph/pull/1527/files#r749653226

The idea is that `CpgPass` subtype can parametrize whether it accepts EC or not with type parameter `T`

### DIFF
--- a/codepropertygraph/src/main/scala/io/shiftleft/passes/CpgPass.scala
+++ b/codepropertygraph/src/main/scala/io/shiftleft/passes/CpgPass.scala
@@ -12,6 +12,24 @@ import scala.annotation.nowarn
 import scala.concurrent.ExecutionContext
 import scala.concurrent.duration.DurationLong
 
+trait MaybeExecutionContext[T] {
+  type ExecCtx
+  def value: ExecCtx
+}
+
+object MaybeExecutionContext {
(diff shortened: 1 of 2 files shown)


codepropertygraph/src/main/scala/io/shiftleft/passes/CpgPass.scala
```diff
@@ -12,6 +12,24 @@ import scala.annotation.nowarn
 import scala.concurrent.ExecutionContext
 import scala.concurrent.duration.DurationLong
 
+trait MaybeExecutionContext[T] {
```
note, line 15: This typeclass in needed only to be able to constraint type parameter `T` of `CpgPass`. The alternative would be to use unconstrained `T` - it would be less readable but with some comments it may be also fine

note:
OK, I gave up on this approach due to `def passes(cpg: Cpg): Iterator[CpgPassBase]`.

It seemed to me that:

```
trait NoExecutionContextRequired extends ExecutionContext
```

Together with making `MaybeExecutionContext` and `CpgPassBase` covariant may work but `ec` actually appears in contravariant position...

PR #1532, opened 2021-11-23
Alternative proposal for propagating MDC

bbrehm:
An alternative proposal for handling the MDC propagation, aka https://github.com/ShiftLeftSecurity/product/issues/8624

We can simplify the entire thing by only running cpg analysis in MDC-specific forkjoin threadpools. Then we  can summon the pool again from `Thread.currentThread()` and use it as implicit for scala-based parallelism; java stream-based parallelism will stay automatically stay inside the pool; in the rare cases that we need explicit worker-threads, we can set their MDC by hand. Will link related codescience PR.


ml86:
I would prefer a solution where the user does not need to be aware of MDC.

bbrehm:
The idea is that you only need to be aware of MDC when setting up MDC for a new cpg analysis, or when using `new Thread`. Otherwise you only need to avoid explicit/implicit use of the global execution-context -- instead of `scala.concurrent.ExecutionContext.global`, you need to use `io.shiftleft.ExecutionContextSummoning.summonExecutionContext`. If using implicits, instead of `import scala.concurrent.ExecutionContext.Implicits._`, set `implicit val ec:ExecutionContext = `io.shiftleft.ExecutionContextSummoning.summonExecutionContext`. Java parallel stream automatically picks up the right context.

codepropertygraph/src/main/scala/io/shiftleft/passes/ParallelCpgPass.scala
```diff
@@ -232,12 +237,17 @@ abstract class ConcurrentWriterCpgPass[T <: AnyRef](cpg: Cpg, outName: String =
...
+                       prefix: String,
+                       inverse: Boolean,
+                       mdc: java.util.Map[String, String] = MDC.getCopyOfContextMap)
+      extends Runnable {
 
     val queue = new LinkedBlockingQueue[Option[DiffGraph]](ConcurrentWriterCpgPass.writerQueueCapacity)
 
     override def run(): Unit = {
       try {
+        MDC.setContextMap(mdc)
```
ml86, line 250: Did we not set this in line 40?

bbrehm: It may look like it, but their are two entirely different inner `private class Writer` :(

codepropertygraph/src/main/scala/io/shiftleft/passes/ParallelCpgPass.scala
```diff
@@ -33,7 +36,8 @@ abstract class ParallelCpgPass[T](cpg: Cpg, outName: String = "", keyPools: Opti
                             prefix: String = "",
                             inverse: Boolean = false)(f: Writer => Unit): Unit = {
     val writer = new Writer(serializedCpg, prefix, inverse)
-    val writerThread = new Thread(writer)
+    val mdc = MDC.getCopyOfContextMap
+    val writerThread = new Thread({ MDC.setContextMap(mdc); writer })
```
ml86, line 40: Can we wrap this in a `createThread` method?

bbrehm: We could. But this logic is needed in 2 places (one of which is legacy and due for deprecation) in one file and is < 1 extra line.

Therefore it makes no sense to abstract this.

bbrehm: I moved the logic around so that both Writer classes use the same.

codepropertygraph/src/main/scala/io/shiftleft/passes/ParallelCpgPass.scala
```diff
@@ -232,12 +241,17 @@ abstract class ConcurrentWriterCpgPass[T <: AnyRef](cpg: Cpg, outName: String =
     }
   }
 
-  private class Writer(serializedCpg: SerializedCpg, prefix: String, inverse: Boolean) extends Runnable {
+  private class Writer(serializedCpg: SerializedCpg,
+                       prefix: String,
+                       inverse: Boolean,
+                       mdc: java.util.Map[String, String] = MDC.getCopyOfContextMap)
```
note, line 247: Nitpick:

For sake of consistence I would make `mdc` look the same for both `Writer`s. My personal preference would be to remove the default value here and provide it explicitly as we do in another `Writer`

codepropertygraph/src/main/scala/io/shiftleft/utils/ExecutionContextSummoning.scala
```diff
@@ -0,0 +1,135 @@
...
+          case ec: ExecutionContext => ec
+          case _                    => ExecutionContext.global
+        }
+      case _ => ExecutionContext.global
+    }
+  }
+
+  val logger = LoggerFactory.getLogger(getClass)
+
+  //adapted from  scala/concurrent/impl/ExecutionContextImpl.scala
```
note, line 22: How about moving this comment right above `MDCThreadFactory`? For me it wasn't obvious that the comment referred to `MDCThreadFactory`. I would also add github link to the code, I believe it is [MDCThreadFactory](https://github.com/scala/scala/blob/2.13.x/src/library/scala/concurrent/impl/ExecutionContextImpl.scala#L27) ?

codepropertygraph/src/main/scala/io/shiftleft/passes/ParallelIteratorExecutor.scala
```diff
@@ -1,10 +1,9 @@
 package io.shiftleft.passes
 
-import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.duration.Duration
-import scala.concurrent.{Await, Future}
+import scala.concurrent.{Await, ExecutionContext, Future}
 
-class ParallelIteratorExecutor[T](iterator: Iterator[T]) {
+class ParallelIteratorExecutor[T](iterator: Iterator[T])(implicit val executionContext: ExecutionContext) {
```
note, line 6: Nitpick: IMO `implicit executionContext`, namely without `val`, would be more accurate as we don't expect anything to access this value

note:
Apart from nitpicks it looks good to me but leaving out approval to someone more familiar with our code base and with problem at hand

ml86 (approved):

PR #1533, opened 2021-11-26
Added DATA_FLOW edge to the schema

DavidBakerEffendi:
- I am experimenting to see if `reachableBy` results can be stored in the CPG and used in later increments.
- This is [being done in Plume](https://github.com/plume-oss/plume/blob/0b0bb620b0411e171c9af2ed21262f7a5d318b52/src/main/scala/io/github/plume/oss/drivers/OverflowDbDriver.scala#L215) to find out if one can save the data-flow paths from these for later use e.g. on code change only update CPG on changed classes and re-use old results.
- Part of this is seeing if there is a negligible performance tradeoff from storing and if this speeds up data-flow analysis long term.

TODO: Is this in the right place? I know there is the `Hidden` schema file but not sure how to access the `cfgNode` from there.

### DIFF
--- a/schema/src/main/scala/io/shiftleft/codepropertygraph/schema/Cfg.scala
+++ b/schema/src/main/scala/io/shiftleft/codepropertygraph/schema/Cfg.scala
@@ -90,18 +90,17 @@ object Cfg extends SchemaBase {
     methodRef.addOutEdge(edge = cfg, inNode = methodReturn)
     typeRef.addOutEdge(edge = cfg, inNode = methodReturn)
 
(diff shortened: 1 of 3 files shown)


DavidBakerEffendi:
@mpollmeier mind giving this a look too? Just want to make sure everything is in the right place and I also need a protoId

mpollmeier:
I'll let @fabsx00 judge whether this is the best place, but I can confirm that you've done the right thing with the proto id - pick a number and 🤞🏻 

Context: the codegen is validating that the ids are unique. We have some closed-source schema extensions, but so far no one was keen enough to pick `12345`, so this should be fine :)

fabsx00 (approved):

PR #1549, opened 2021-12-07
Enable Scala 3 cross build.

ml86:


build.sbt
```diff
@@ -6,7 +6,8 @@ val overflowdbVersion = "1.83"
 inThisBuild(
   List(
     organization := "io.shiftleft",
-    scalaVersion := "2.13.7",
+    scalaVersion := "3.1.0",
```
mpollmeier, line 9: once merged, can you inform others that they should use `+publishM2 / +publishLocal` for local dev?
downstream repos are most likely still on scala 2...

mpollmeier (approved):

PR #1575, opened 2022-01-06
Add operators to replace magic strings in javasrc2cpg

johannescoetzee:
* Add `<operator>.arrayCreator` for calls to array creation expressions such as `new int[]`. Alternatively, we can re-use `alloc` for this.

* Add `<operator>.alloc` for object instantiation (this is called before `<init>`.


fabsx00 (approved):
Since this just replaces the magic strings, it makes visible how things currently are, so, let's get this merged first. Seems to me though that we have to take a step back soon to see how we want to model allocation/initialization across languages.

fabsx00:
As discussed on slack, holding off on merging because these two operators are only used in javasrc, so, we might as well figure out how we want to represent this across frontends before making changes.

johannescoetzee:
Closing this until we've decided which of these operators to keep

PR #1578, opened 2022-01-07
cleanup imports, mostly to fix intellij squiggles

mpollmeier:
@mal-tee triggered an investigation that led us to figure out that some 
implicit imports break intellij's scalac presentation compiler, 
specifically `import overflowdb.traversal.iterableToTraversal`

While this was never an issue for scalac, it's straightforward to fix and
gives everyone a better experience with intellij.

https://discord.com/channels/832209896089976854/832214243230744626/928317907576946739 
https://github.com/mal-tee/intellij-cpg-problems-example

before:
before

after:
after


fabsx00 (approved):
Good that it removes the squiggles, not so good that we need to specify those three classes manually...

mpollmeier:
it'll all change with scala 3...

Issue #1577, opened 2022-01-06
Errors during createAndApply of ParallelCpgPass not escalated


simkoc:
When I try to create edges that violate the CPG spec an error is thrown internally but not escalated in a way that I am able to handle it.

[An example for a succeeding test even with a violation would be](https://github.com/simkoc/cpg-bug-poc/tree/issue2):
```
// [METHOD] -> POST_DOMINATE -> [ BLOCK ]

class CreateInitialSetupPass(something: List[Int],
                             cpg: Cpg,
                             keyPool: IntervalKeyPool)
    extends ParallelCpgPass[Int](
      cpg,
      keyPools = Some(keyPool.split(something.size))) {

  override def partIterator: Iterator[Int] = something.iterator

  override def runOnPart(part: Int): Iterator[DiffGraph] = {
    implicit val diffGraph: DiffGraph.Builder = DiffGraph.newBuilder
    val method = nodes.NewMethod().name("dlr_main()")
    diffGraph.addNode(method)
    val block = nodes.NewBlock().code("DLR MAIN BLOCK")
    diffGraph.addNode(block)
    diffGraph.addEdge(method, block, EdgeTypes.POST_DOMINATE)
    Iterator(diffGraph.build())
  }
}

class TestForBug  extends AnyWordSpec with Matchers {

  "the cpg creation test" should {
    "fail" in {
      val cpg = X2Cpg.newEmptyCpg(Some("/tmp/out.cpg"))
      val keyPools = KeyPoolCreator.obtain(1).iterator
      val run = new CreateInitialSetupPass(List(0), cpg, keyPools.next())
      run.createAndApply()
    }
  }

}
```

The test succeeds even though the creation fails (`sbt test`). There is a stacktrace printed in the shell:

```
Exception in thread "Writer" java.lang.RuntimeException: Edge with type='POST_DOMINATE' with direction='OUT' not supported by nodeType='METHOD'
	at overflowdb.NodeDb.storeAdjacentNode(NodeDb.java:637)
	at overflowdb.NodeDb.storeAdjacentNode(NodeDb.java:616)
	at overflowdb.NodeDb.addEdge(NodeDb.java:303)
	at overflowdb.NodeRef.addEdge(NodeRef.java:160)
	at overflowdb.SemiEdge.$minus$minus$greater(SyntacticSugar.scala:59)
	at io.shiftleft.passes.DiffGraph$Applier.odbAddEdge(DiffGraph.scala:411)
[...]

PR #1585, opened 2022-01-19
catch and rethrow exceptions in ParallelCpgPass

mpollmeier:
as discussed in https://github.com/ShiftLeftSecurity/codepropertygraph/issues/1577

### DIFF
--- a/codepropertygraph/src/main/scala/io/shiftleft/passes/ParallelCpgPass.scala
+++ b/codepropertygraph/src/main/scala/io/shiftleft/passes/ParallelCpgPass.scala
@@ -38,6 +38,10 @@ abstract class ParallelCpgPass[T](cpg: Cpg, outName: String = "", keyPools: Opti
     val writer = new Writer(serializedCpg, prefix, inverse, MDC.getCopyOfContextMap())
     val writerThread = new Thread(writer)
     writerThread.setName("Writer")
+    var exceptionCaught: Option[Throwable] = None
+    writerThread.setUncaughtExceptionHandler { (_: Thread, t: Throwable) =>
+      exceptionCaught = Option(t)
+    }
     writerThread.start()
     try {
       f(writer)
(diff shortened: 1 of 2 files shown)


codepropertygraph/src/main/scala/io/shiftleft/passes/ParallelCpgPass.scala
```diff
@@ -38,6 +38,10 @@ abstract class ParallelCpgPass[T](cpg: Cpg, outName: String = "", keyPools: Opti
     val writer = new Writer(serializedCpg, prefix, inverse, MDC.getCopyOfContextMap())
     val writerThread = new Thread(writer)
     writerThread.setName("Writer")
+    var exceptionCaught: Option[Throwable] = None
```
bbrehm, line 41: This is ok because `writerThread.join()` is a memory barrier; `@volatile` would come at no perceptible performance cost and make it "more obviously OK".

Feel free to disregard the suggestion if you disagree, that's a pure coding style question.

codepropertygraph/src/main/scala/io/shiftleft/passes/ParallelCpgPass.scala
```diff
@@ -48,6 +52,11 @@ abstract class ParallelCpgPass[T](cpg: Cpg, outName: String = "", keyPools: Opti
       writer.enqueue(None, None)
       writerThread.join()
     }
+
+    exceptionCaught.foreach { exception =>
+      baseLogger.warn(s"Exception in parallel CPG pass $name:", exception)
```
bbrehm, line 57: The exception is already logged in line 50.

bbrehm (approved):

codepropertygraph/src/main/scala/io/shiftleft/passes/ParallelCpgPass.scala
mpollmeier: No it's not... that's what I first thought as well, but because we start a new Thread (line 39), and the exception is thrown in that thread (and handled by the newly registered `UncaughtExceptionHandler`)

PR #1590, opened 2022-02-08
adapt to odb batched update

bbrehm:


codepropertygraph/src/main/scala/io/shiftleft/passes/ParallelCpgPass.scala
```diff
@@ -256,10 +271,14 @@ abstract class ConcurrentWriterCpgPass[T <: AnyRef](cpg: Cpg, outName: String =
                        mdc: java.util.Map[String, String])
       extends Runnable {
 
-    val queue = new LinkedBlockingQueue[Option[DiffGraph]](ConcurrentWriterCpgPass.writerQueueCapacity)
+    val queue =
+      new LinkedBlockingQueue[Option[overflowdb.BatchedUpdate.DiffGraph]](ConcurrentWriterCpgPass.writerQueueCapacity)
+
+    @volatile var raisedException: Exception = null
```
mpollmeier, line 277:
```suggestion
    @volatile private var raisedException: Exception = null
```

semanticcpg/src/test/scala/io/shiftleft/semanticcpg/passes/CfgDominatorFrontierTests.scala
```diff
@@ -23,7 +23,7 @@ class CfgDominatorFrontierTests extends AnyWordSpec with Matchers {
       node.in("CFG").asScala
   }
 
-  private class TestDomTreeAdapter(immediateDominators: Map[Node, Node]) extends DomTreeAdapter[Node] {
+  private class TestDomTreeAdapter(immediateDominators: scala.collection.Map[Node, Node]) extends DomTreeAdapter[Node] {
```
mpollmeier, line 26: was that change intended?

mpollmeier (approved):

codepropertygraph/src/main/scala/io/shiftleft/passes/ParallelCpgPass.scala
```diff
@@ -271,17 +290,33 @@ abstract class ConcurrentWriterCpgPass[T <: AnyRef](cpg: Cpg, outName: String =
...
-              if (doSerialize) {
-                store(serialize(appliedDiffGraph, withInverse),
-                      generateOutFileName(prefix, outName, index),
+              val listener =
+                if (withInverse) new BatchUpdateInverseListener
+                else if (doSerialize) new BatchUpdateForwardListener
+                else null
+
+              nDiffT += overflowdb.BatchedUpdate
+                .applyDiff(cpg.graph, diffGraph, keyPool.getOrElse(null), listener)
```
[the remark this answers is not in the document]
bbrehm: If this was scala, I'd agree. The new BatchedUpdate API in odb is java, though, where `null` is the normal way of encoding optional non-primitive values.

semanticcpg/src/test/scala/io/shiftleft/semanticcpg/passes/CfgDominatorFrontierTests.scala
bbrehm: Yes. `Map` by default refers to `scala.collection.immutable.Map`, whereas `scala.collection.Map` is a shared super of both mutable and immutable maps. I consider that design decision of the scala team rather questionable.

In order to ensure determinism, I switched the actual map out to `scala.collection.mutable.LinkedHashMap`, so this change was necessary if we want to avoid unnecessary copies.

codepropertygraph/src/main/scala/io/shiftleft/passes/ParallelCpgPass.scala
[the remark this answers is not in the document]
mpollmeier: java has `java.util.Optional` since java8, and we use it in many places already. it just helps to clarify your intent: this thing is optional. 
if you really don't like it, then how about the secondary constructor idea?

semanticcpg/src/test/scala/io/shiftleft/semanticcpg/passes/CfgDominatorFrontierTests.scala
mpollmeier: 👍🏻

PR #1611, opened 2022-03-18
Some more utilities for diffgraph/batchedUpdate migration

bbrehm:
Add conversion from legacy diffgraph -> new diffgraph API and restructures the new-style cpgpass hierarchy. Furthermore, it adds an API to CpgPassBase to genererate a batchedUpdate-style diff.

The conversion of diffgraphs allows us to indefinitely delay migration of passes.

The new API to generate new-style diffs from a pass, together with the logging facilities, allows us to create "compound passes" that collect diffs from many different passes, and applies them together. This is exactly how the framework-passes are run in codescience (PR will be linked).

The restructuring of the class hierarchy, i.e. the new shared super of ForkJoinParallelCpgPass and ConcurrentWriterCpgPass, allows us to share some code and allows users to look into the internals more easily -- the API between both is intentionally identical, so let's formalize that fact.

I am very open to naming and API discussions and especially suggestions on that front -- my names for the new stuff suck.

### DIFF
--- a/codepropertygraph/src/main/scala/io/shiftleft/passes/CpgPass.scala
+++ b/codepropertygraph/src/main/scala/io/shiftleft/passes/CpgPass.scala
@@ -85,7 +85,7 @@ abstract class CpgPass(cpg: Cpg, outName: String = "", keyPool: Option[KeyPool] 
 
   override def runWithBuilder(builder: BatchedUpdate.DiffGraphBuilder): Int = {
     var nParts = 0
-    for(diff <- run()){
+    for (diff <- run()) {
       diff.convertToOdbStyle(cpg, builder)
       nParts += 1
     }
@@ -144,13 +144,13 @@ abstract class SimpleCpgPass(cpg: Cpg, outName: String = "", keyPool: Option[Key
  * passes eagerly, and releases them only when the entire chain has run.
  * */
 abstract class ForkJoinParallelCpgPass[T <: AnyRef](cpg: Cpg, outName: String = "", keyPool: Option[KeyPool] = None)
-  extends NewStyleCpgPassBase[T]{
+    extends NewStyleCpgPassBase[T] {
 
   override def createApplySerializeAndStore(
-                                             serializedCpg: SerializedCpg,
-                                             inverse: Boolean = false,
-                                             prefix: String = ""
-                                           ): Unit = {
+    serializedCpg: SerializedCpg,
+    inverse: Boolean = false,
+    prefix: String = ""
+  ): Unit = {
     baseLogger.info(s"Start of pass: $name")
     val nanosStart = System.nanoTime()
     var nParts     = 0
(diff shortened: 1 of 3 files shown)


codepropertygraph/src/main/scala/io/shiftleft/passes/CpgPass.scala
```diff
@@ -81,6 +83,15 @@ abstract class CpgPass(cpg: Cpg, outName: String = "", keyPool: Option[KeyPool]
     }
   }
 
+  override def runWithBuilder(builder: BatchedUpdate.DiffGraphBuilder): Int = {
```
fabsx00, line 86: Can you add a comment here that describes what the method does and what it returns?

codepropertygraph/src/main/scala/io/shiftleft/passes/CpgPass.scala
```diff
@@ -222,16 +195,74 @@ abstract class ForkJoinParallelCpgPass[T <: AnyRef](cpg: Cpg, outName: String =
...
+  *
+  * Please don't subclass this directly. The only reason it's not sealed is that this would mess with our file
+  * hierarchy.
+  */
+abstract class NewStyleCpgPassBase[T <: AnyRef] extends CpgPassBase {
+  type DiffGraphBuilder = overflowdb.BatchedUpdate.DiffGraphBuilder
+  // generate Array of parts that can be processed in parallel
+  def generateParts(): Array[_ <: AnyRef]
+  // setup large data structures, acquire external resources
+  def init(): Unit = {}
```
fabsx00, line 223: If these need to be define by the child class, then we might just want to leave them empty, as in `def init(): Unit`. That way, the IDE will automatically define them for child classes.

fabsx00 (approved):
The one thing I'm worried about is that delaying indefinitely means that we'll probably not do it. Are we still getting deprecation warnings though?

codepropertygraph/src/main/scala/io/shiftleft/passes/CpgPass.scala
bbrehm: `init()` and `finish()` were taken from the now-legacy CpgPass interface -- so you tell me how you intended them to be used ;)

The idea is that a pass may want to run some initialization and cleanup code immediately before and after running. However, most passes don't want to do that, so we do the usual OOP dance of providing a hook in the clazz's vtable that we fill with a NOP default.

Example for `finish()` is closing of file descriptors and clearing pointers to large arrays; example for `init()` is allocation of large arrays and setup code that needs to run serially before the parallel section. Yeah, enterprising individuals could put that into `generateParts()` but better provide a standard way with appropriate guarantees.

The large-array (or generally large-structure) stuff is important because of the way we organize our LayerCreator things: First construct all passes, then run them, and only then drop all references to the passes, allowing them to be garbage collected. Hence, internal shared data structures for parallel passes have an overly large lifetime unless cleared by hand, and hence we want to provide a hook for clearing that stuff.

bbrehm:
Yes, we are still getting deprecation warnings. And I am continuing to work on the migration.

I did not originally want to add this kind of compat code, but:
1. it turned out to be much simpler than expected
2. it turned out to be more or less necessary in order to permit gradual migration of framework passes with the multistageRunner in codescience
3. it is a big productivity plus -- this means that odb2 progress is no longer blocked by the ongoing cpg pass migration
4. it is pretty always nice to be able to retain more backwards compatibility for longer time-spans than expected!

After seeing how simple this PR ended up, I feel somewhat silly about allowing the CpgPass migration to block odb2 progress in the first place...

codepropertygraph/src/main/scala/io/shiftleft/passes/CpgPass.scala
bbrehm: Added a comment to the API definition in CpgPassBase

ml86 (approved):

PR #1624, opened 2022-04-05
Rename `PARAMETER_INDEX` to `INDEX`

fabsx00:
This is a follow-up for PR #1622. As I started using `parameterIndex`, I noticed that `parameter.parameterIndex` is redundant. In fact, this is the same index that we use for `argumentIndex` - also a rather long name. My short term suggestion is that it's `parameter.index`. The long term suggestion is that we also replace `argumentIndex` with `index`, so that we end up with `parameter.index` and `argument.index`. Notice that the same rules apply for these two indices as we need to match them.

### DIFF
--- a/schema/src/main/scala/io/shiftleft/codepropertygraph/schema/Method.scala
+++ b/schema/src/main/scala/io/shiftleft/codepropertygraph/schema/Method.scala
@@ -1,6 +1,5 @@
 package io.shiftleft.codepropertygraph.schema
 
-import io.shiftleft.codepropertygraph.schema.CpgSchema.PropertyDefaults
 import overflowdb.schema.Property.ValueType
 import overflowdb.schema.{NodeType, SchemaBuilder, SchemaInfo}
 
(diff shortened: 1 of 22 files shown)


schema/src/main/scala/io/shiftleft/codepropertygraph/schema/Method.scala
```diff
@@ -86,17 +85,6 @@ object Method extends SchemaBase {
...
-      .addProperty(
-        name = "PARAMETER_INDEX",
-        valueType = ValueType.Int,
-        comment = """
-            |Specifies the parameter's number. For the `this` parameter, the parameter index `0` is used.
-            |""".stripMargin
-      )
-      .mandatory(PropertyDefaults.Int)
-      .protoId(222)
-
```
ml86, line 99: Removing this would mean all frontends need to be updated in one go or are we currently just ignoring unknown properties while loading?

fabsx00: I'm removing this property, which I introduced yesterday and which isn't used yet.

ml86 (approved):

schema/src/main/scala/io/shiftleft/codepropertygraph/schema/Method.scala
ml86: Correct. Can be removed.

PR #1641, opened 2022-06-20
allow to tag metaData node

bbrehm:
As discussed in https://github.com/ShiftLeftSecurity/codescience/pull/6191 and linked issues, we need a way to stash some global data about a cpg, in ad-hoc internal structures.

One way we came up with is to attach TAG nodes to the metadata node. This PR permits to tag metadata nodes.

This part of the feature is very much open to bike-shedding. The only requirements are:

1. Don't clutter the graph schema
2. Don't overload the FINDING node even further
3. Don't attempt to spec the actual data format or contents. People over in teams CS/UI want space to experiment now; after customer feedback is in, we can then eventually figure out a schema for whatever this turns out to be used for. Until then, we just need a dumpster for ad-hoc structured blobs.

The question of where to place such stuff in the graph is mostly independent from the other considerations (passes that produce this data during scans, binary protocol + proto hackage to transmit this, handling in the web-services side, slick UI that displays this data).

### DIFF
--- a/schema/src/main/scala/io/shiftleft/codepropertygraph/schema/CpgSchema.scala
+++ b/schema/src/main/scala/io/shiftleft/codepropertygraph/schema/CpgSchema.scala
@@ -24,7 +24,7 @@ class CpgSchema(builder: SchemaBuilder) {
   val shortcuts = Shortcuts(builder, base, method, ast, typeSchema, fs)
 
   val sourceSpecific  = Comment(builder, ast, fs)
-  val tagsAndLocation = TagsAndLocation(builder, base, typeSchema, method, ast, fs, callGraph)
+  val tagsAndLocation = TagsAndLocation(builder, base, typeSchema, method, ast, fs, callGraph, metaData)
   val binding         = Binding(builder, base, typeSchema, method, callGraph)
   val annotation      = Annotation(builder, base, method, typeSchema, ast, shortcuts)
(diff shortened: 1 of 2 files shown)


mpollmeier:
> very much open to bike-shedding

I don't have the complete context of the linked issue, but I'm wondering: why are these tags linked to the (one and only) metadata node? Instead, they could just be free-hanging TAG nodes...

bbrehm:
My reason for attaching to metadata is that there are a lot of TAG nodes in the graph, and grabbing the global dumpster-data should not require us to iterate over all of them. 

On the other hand, sure, why not -- we won't do that inside of a hot loop.

My tendency is to write all code with performance by default, and only permit bad scaling / complexity class after discussion (as opposed to the philosophy of writing everything "idiomatically" and only considering performance after discussion / profiling).

In this specific case, I actually agree with your comment -- free floating is easier on the schema and fast enough for our needs.

I'll keep this open for a couple days in case that somebody else chimes in, but otherwise you convinced me.

mpollmeier:
OTOH, if we go for free-hanging TAG nodes, then there's nothing to move, really :)

ml86:
I am against free floating TAG nodes. After all a TAG node is supposed to "tag" something...
I suggest we just allow tagging on all node types. That is generic enough and we do not need to move something to CS schema.

PR #1660, opened 2022-10-05
Improve documentation for FULL_NAME property on ANNOTATION nodes.

ml86:
I hope this clarifies the missunderstandings discussed in:
https://github.com/ShiftLeftSecurity/codescience/pull/6517


bbrehm (approved):
I still prefer something like https://github.com/ShiftLeftSecurity/codepropertygraph/pull/1661 but no need to block this definite improvement

fabsx00 (approved):

PR #1661, opened 2022-10-05
Alternative improved documentation for FULL_NAME property on ANNOTATION nodes.

bbrehm:
Alternative to https://github.com/ShiftLeftSecurity/codepropertygraph/pull/1660

I wrote it at the same time, but @ml86 was faster with opening the PR ;)


fabsx00 (changes requested):
This would go into the official documentation.

ml86:
The alternative was already merged.

PR #1662, opened 2022-10-11
fix bug in diffgraph proto forward serialization

bbrehm:
@claudiuvursache uncovered a logic bug in the proto forward serialization.

The fact that it took until now to discover that problem suggests that this feature is quite sparsely used.


claudiuvursache (approved):

claudiuvursache:
:raised_hands:

bbrehm:
I'll remark that unit tests are missing for afaiu 100% of the forward serialization logic. Well, this PR at least cannot make anything worse ;)

(I think the serialization-of-diffgraph feature should rather be removed than us starting to write proper tests for it)

PR #1672, opened 2022-10-18
Remove unused logic

claudiuvursache:


bbrehm (approved):
There is lots of more unused logic in there. You will remove that in a follow-up?

Also, let's think a little about the commit message. We're removing a feature (overlay loading) that we will probably need to occasionally access in the future, for working with even older CPGs. 

That is OK -- we can just check out the previous commit of codepropertygraph / joern / codescience / ocular in order to handle that CPG. But for that, we need to find the correct commit, i.e. we need to anticipate what our future selves will grep for in the commit logs. I am thinking something like

```
Removes loading of overlay protos (serialized diffgraph) in  .bin.zip format. For handling such old files, check out the previous commit.
 ```

but feel free to add more greppable keywords.

PR #1676, opened 2022-10-25
Add yield control structure type

johannescoetzee:
I'm not sure if controlstructure is a good fit for yield (since yield is closer to return that we have a specific node type for), but it seems like the best option for it that doesn't require new node types.


fabsx00 (approved):
Let's go with it for now.

PR #1678, opened 2022-11-02
Introduce `type CpgPass = SimpleCpgPass`

claudiuvursache:
```
codepropertygraph/src/main/scala/io/shiftleft/passes/CpgPass.scala
13:/* SimpleCpgPass is a replacement for CpgPass.
19: * The subtleties are gone with SimpleCpgPass.
21: * Note that SimpleCpgPass does not support lazy evaluation to reduce peak memory consumption. Take care before porting passes that
49:type CpgPass = SimpleCpgPass
51:abstract class SimpleCpgPass(cpg: Cpg, outName: String = "", keyPool: Option[KeyPool] = None)
```


mpollmeier (approved):
do you want to do this refactor in small steps, or why not rename SimplePass to CpgPass?

bbrehm:
If you want more compat, you could add `@deprecated class SimpleCpgPass extends CpgPass` -- that way everything will continue to compile, albeit with warnings (in case someone has scripts lying around that you won't find for fixing the renaming).

bbrehm (approved):

PR #1679, opened 2022-11-02
Remove `DiffGraphProtoSerializer`

fabsx00:
We removed the undo feature, yet, this component seemed to still be used and that lead to performance degradation. Removing this now, so that it may never be used by accident.


ml86 (approved):

mpollmeier:
For reference: when importing https://github.com/keycloak/keycloak into joern with `Xmx12g`, `overflowdb.Graph` overflows to disk, taking up only 1% of heap at the end. At the same time `BatchUpdateInverseListener` grew unbounded and makes us run into an OOMError.

fabsx00:
Yep, I'm quite sure this is what we were looking for.

PR #1680, opened 2022-11-03
Schema: VersionControl Node for VCS Information if Present

DavidBakerEffendi:
Added a version control node that describes the following properties:
- VCS_SYSTEM: E.g. Git or SVN
- REMOTE (optional): The remote repository URI
- BRANCH: The branch this code is on
- AUTHOR: Who created the revision
- REVISION_ID: The ID of the code revision, e.g. commit hash
- PARENT_REVISION_IDS: A list of direct parent revision IDs
- REVISION_MESSAGE: The accompanying message of the revision
- MODIFIED_FILES: A list of changed files
- REVISION_TAGS: A list of tags on this commit (no tags from forward commits)

### DIFF
--- a/schema/src/main/scala/io/shiftleft/codepropertygraph/schema/VersionControl.scala
+++ b/schema/src/main/scala/io/shiftleft/codepropertygraph/schema/VersionControl.scala
@@ -1,0 +1,100 @@
+package io.shiftleft.codepropertygraph.schema
+
+import overflowdb.schema.{NodeType, SchemaBuilder, SchemaInfo}
+
+object VersionControl extends SchemaBase {
+
+  override def docIndex: Int = 21
+
+  override def description: String =
+    """
+      |Version control systems (VCS) are a natural part of software development and, as such, the code property graph
+      |will encode some key details about the VCS in use, if any. These details can help correlate the code version
+      |corresponding to the generated CPG. One can choose to explore the development evolution of a versioned codebase
+      |using the code property graph.
+      |""".stripMargin
+
+  def apply(builder: SchemaBuilder) = new Schema(builder)
+
+  class Schema(builder: SchemaBuilder) {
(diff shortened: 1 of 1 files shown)


DavidBakerEffendi:
Truth be told, `git-lfs` is not working on my M1 Mac with `sbt`, so I cannot reliably test this locally 😅

DavidBakerEffendi:
An implementation that can extract these details are present on [dave/detect-git](https://github.com/joernio/joern/blob/dave/detect-git/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/base/VersionControlPass.scala)

ml86:
Hmm, to be honest i think representing VCS information in the CPG is out of the current scope of information CPGs represent. So far we have been including things like dependency information which is also not code but the dependencies together with the analysed code form a union where you sometimes can only judge about security problems if you have both. 
VCS information on the other hand is just used to represent a single point in a projects code history but from the analysis point of view we are anyway only looking at a single snapshot in that timeline. 
Only if we would want to start representing diffs between commits we would need that kind of information but most likely the proposed representation would not suffice since it is not associated to individual nodes.

In the end people scratching their head because they dont remember which exact version of the code they analysed is not really our business. They could just have put the commit tag in the CPG name.

If you and others still feel this is a great feature i would suggest a simple VSC string property on the META_DATA node where we could just mangle all relevant information by concatenation. E.g. `VSC = "GIT:<branchName>:<author>..."` This would avoid the problem that different VCS have complete different operation semantics and thus defining one group of properties which can be meaningfully filled by all of them wont be possible. And starting to define groups of properties for each different VCS we want to support sounds like overkill.

DavidBakerEffendi:
I think best is to shelve this and keep the work available in case anyone wants to see an example of this for a schema extension.

PR #1683, opened 2022-11-11
Move depth first order numbering of CFG nodes to codepropertygraph repo.

ml86:
This is a preparation step to avoid calculation the depth first order
numbering multiple times.


fabsx00:
Ok, let's merge it once we need it.

PR #1707, opened 2023-05-26
check in generated classes

mpollmeier:
* allows users/developers to validate the changes in the generated files when they make a change to the schema or codegen
* allows us to easily compare them in the history
* users trying to find sources (e.g. for their own understanding) can now just search in the repo, e.g. on github, without having to build them locally
* depends on https://github.com/ShiftLeftSecurity/overflowdb-codegen/pull/215

n.b. after this is merged I'll add some more functionality to the codegen sbt plugin to 
a) allow to disable it (so that one can manually fiddle with the domain classes) and
b) run a generateDomainClassesCheck (e.g. on PR), similar to `scalafmtCheck` to ensure that the checked in sources always correspond with the latest schema and codegen


mpollmeier:
will greenify after https://github.com/ShiftLeftSecurity/codepropertygraph/pull/1708 is in

PR #1737, opened 2023-10-17
FILE_CONENT property proposal.

ml86:


fabsx00 (changes requested):
Can we just call them `CONTENT`, `OFFSET`, ...?

ml86:
We already have a CONTENT property which is used for CONFIG_FILE nodes. I did not want to reuse it to avoid confusion. As for the offsets: Yes, i can just name them OFFSET end OFFSET_END.

ml86:
Ok, i now reuse the already existing CONTENT property.

fabsx00 (approved):

johannescoetzee (approved):

PR #1743, opened 2023-11-09
fail if repo is dirty: use 3rd party action

mpollmeier:
tested via https://github.com/ShiftLeftSecurity/codepropertygraph/pull/1742
grafik


mpollmeier:
oh, actually this never works... the action isn't published apparently

max-leuthaeuser (approved):

PR #1759, opened 2024-02-29
Performance matrix logger

pandurangpatil:
Added one utility to record the performance matrix in the following format. 

1. One if you `initialize` method and set the file where to write the matrix in csv format then it will start writing to that specific file with the configured interval of time.

DateTime | MainStage (group of passes like overlay pass) | SubStage (Pass to be executed) | MetaData (some additional data like which CPG pass is being used) | CPU (%) | Used Memory (MB) | Max Heap (MB)
-- | -- | -- | -- | -- | -- | --

2. It will also log debug logs when a pass is invoked and when the pass is ended with the time it took to finish the pass.

Sample report 
[performancematrix.csv](https://github.com/ShiftLeftSecurity/codepropertygraph/files/14521184/performancematrix.csv)

### DIFF
--- a/codepropertygraph/src/main/scala/io/shiftleft/passes/CpgPass.scala
+++ b/codepropertygraph/src/main/scala/io/shiftleft/passes/CpgPass.scala
@@ -67,7 +67,7 @@ abstract class ForkJoinParallelCpgPass[T <: AnyRef](
     prefix: String = ""
   ): Unit = {
     baseLogger.info(s"Start of pass: $name")
-    TimeMetric.initiateNewStage(getClass.getSimpleName, Some(name), "ForkJoinParallelCpgPass")
+    TimeMetric.initiateNewStage(getClass.getSimpleName, Some(name), getClass.getSuperclass.getSimpleName)
     val nanosStart = System.nanoTime()
     var nParts     = 0
     var nanosBuilt = -1L
--- a/codepropertygraph/src/main/scala/io/shiftleft/passes/CpgPass.scala
+++ b/codepropertygraph/src/main/scala/io/shiftleft/passes/CpgPass.scala
@@ -3,6 +3,7 @@ package io.shiftleft.passes
 import com.google.protobuf.GeneratedMessageV3
 import io.shiftleft.SerializedCpg
 import io.shiftleft.codepropertygraph.Cpg
+import io.shiftleft.utils.TimeMetric
 import org.slf4j.{Logger, LoggerFactory, MDC}
 import overflowdb.BatchedUpdate
 
@@ -66,6 +67,7 @@ abstract class ForkJoinParallelCpgPass[T <: AnyRef](
     prefix: String = ""
   ): Unit = {
     baseLogger.info(s"Start of pass: $name")
+    TimeMetric.initiateNewStage(name, "ForkJoinParallelCpgPass")
     val nanosStart = System.nanoTime()
     var nParts     = 0
     var nanosBuilt = -1L
@@ -99,6 +101,7 @@ abstract class ForkJoinParallelCpgPass[T <: AnyRef](
         baseLogger.info(
           f"Pass $name completed in ${(nanosStop - nanosStart) * 1e-6}%.0f ms (${fracRun}%.0f%% on mutations). ${nDiff}%d + ${nDiffT - nDiff}%d changes committed from ${nParts}%d parts.${serializationString}%s"
         )
+        TimeMetric.endLastStage()
       }
     }
   }
--- a/codepropertygraph/src/main/scala/io/shiftleft/passes/CpgPass.scala
+++ b/codepropertygraph/src/main/scala/io/shiftleft/passes/CpgPass.scala
@@ -67,7 +67,7 @@ abstract class ForkJoinParallelCpgPass[T <: AnyRef](
     prefix: String = ""
   ): Unit = {
     baseLogger.info(s"Start of pass: $name")
-    TimeMetric.initiateNewStage(name, "ForkJoinParallelCpgPass")
+    TimeMetric.initiateNewStage(getClass.getSimpleName, Some(name), "ForkJoinParallelCpgPass")
     val nanosStart = System.nanoTime()
     var nParts     = 0
     var nanosBuilt = -1L
--- a/codepropertygraph/src/main/scala/io/shiftleft/passes/ParallelCpgPass.scala
+++ b/codepropertygraph/src/main/scala/io/shiftleft/passes/ParallelCpgPass.scala
@@ -58,7 +58,7 @@ abstract class ConcurrentWriterCpgPass[T <: AnyRef](
   ): Unit = {
     import ConcurrentWriterCpgPass.producerQueueCapacity
     baseLogger.info(s"Start of enhancement: $name")
-    TimeMetric.initiateNewStage(getClass.getSimpleName, Some(name), "ConcurrentWriterCpgPass")
+    TimeMetric.initiateNewStage(getClass.getSimpleName, Some(name), getClass.getSuperclass.getSimpleName)
     val nanosStart = System.nanoTime()
     var nParts     = 0
     var nDiff      = 0
--- a/codepropertygraph/src/main/scala/io/shiftleft/passes/ParallelCpgPass.scala
+++ b/codepropertygraph/src/main/scala/io/shiftleft/passes/ParallelCpgPass.scala
@@ -58,7 +58,7 @@ abstract class ConcurrentWriterCpgPass[T <: AnyRef](
   ): Unit = {
     import ConcurrentWriterCpgPass.producerQueueCapacity
     baseLogger.info(s"Start of enhancement: $name")
-    TimeMetric.initiateNewStage(name, "ConcurrentWriterCpgPass")
+    TimeMetric.initiateNewStage(getClass.getSimpleName, Some(name), "ConcurrentWriterCpgPass")
     val nanosStart = System.nanoTime()
     var nParts     = 0
     var nDiff      = 0
--- a/codepropertygraph/src/main/scala/io/shiftleft/passes/ParallelCpgPass.scala
+++ b/codepropertygraph/src/main/scala/io/shiftleft/passes/ParallelCpgPass.scala
@@ -1,7 +1,7 @@
 package io.shiftleft.passes
 import io.shiftleft.SerializedCpg
 import io.shiftleft.codepropertygraph.Cpg
-import io.shiftleft.utils.ExecutionContextProvider
+import io.shiftleft.utils.{ExecutionContextProvider, TimeMetric}
 import org.slf4j.MDC
 
 import java.util.concurrent.LinkedBlockingQueue
@@ -58,6 +58,7 @@ abstract class ConcurrentWriterCpgPass[T <: AnyRef](
   ): Unit = {
     import ConcurrentWriterCpgPass.producerQueueCapacity
     baseLogger.info(s"Start of enhancement: $name")
+    TimeMetric.initiateNewStage(name, "ConcurrentWriterCpgPass")
     val nanosStart = System.nanoTime()
     var nParts     = 0
     var nDiff      = 0
@@ -124,6 +125,7 @@ abstract class ConcurrentWriterCpgPass[T <: AnyRef](
       baseLogger.info(
         f"Enhancement $name completed in ${(nanosStop - nanosStart) * 1e-6}%.0f ms. ${nDiff}%d  + ${nDiffT - nDiff}%d changes committed from ${nParts}%d parts."
       )
+      TimeMetric.endLastStage()
     }
   }
 
--- a/codepropertygraph/src/main/scala/io/shiftleft/utils/TimeMetric.scala
+++ b/codepropertygraph/src/main/scala/io/shiftleft/utils/TimeMetric.scala
@@ -463,7 +463,7 @@ object TimeMetric {
     * @return
     *   \- Formatted double value in string.
     */
-  private def doubleFormatter(value: Double): String = s"${String.format("%.2f", value)}"
+  private def doubleFormatter(value: Double): String = s"${"%.2f".formatLocal(java.util.Locale.US, value)}"
 
   /** Get current time by initialising internal latest data time tracker.
     * @return
--- a/codepropertygraph/src/main/scala/io/shiftleft/utils/TimeMetric.scala
+++ b/codepropertygraph/src/main/scala/io/shiftleft/utils/TimeMetric.scala
@@ -9,34 +9,108 @@ import java.util
 import java.util.concurrent.ConcurrentLinkedQueue
 import java.util.{Calendar, Date}
 import scala.collection.mutable
+
+/** Stage wise Time, cpu, and memory recording config to pass file path and the recording frequency
+  *
+  * @param resultFile
+  *   \- complete file path where the recording will be saved in csv format.
+  * @param recordFreq
+  *   \- Frequency to write the records in milliseconds.
+  */
 case class TimeMetricRecordConfig(resultFile: String = "cpuandmemoryusage.csv", recordFreq: Int = 60000)
+
+/** These buffers are used for console logs to set the output fix length of the column. This will help make the console
+  * output more readable. This has been kept configurable so that end consumer can adjust according to their use case.
+  * @param MAX_FULLNAME_BUFFER
+  *   \- If the class full names of passes is used to print the stage name on the console. Then this buffer is used
+  *   considering the length the full name of the class
+  * @param MAX_NAME_BUFFER
+  *   \- In case the only simple name of the class is used to print the stage name on the console. Then this buffer is
+  *   used considering the length of the only name of the class.
+  * @param MAX_DONE_TIME_BUFFER
+  *   \- When done status of the stage is updated, it also adds the total time taken by the time and then prints the cpu
+  *   and memory stats afte that. This buffer will make `stats` information to be printed in the same colum for
+  *   `invoked` status line to make it more readable.
+  */
 case class MaxPrintBuffers(MAX_FULLNAME_BUFFER: Int = 90, MAX_NAME_BUFFER: Int = 55, MAX_DONE_TIME_BUFFER: Int = 40)
+
+/** Cpu and Memory stats information to log
+  * @param usedOrAvgMemory
+  *   \- current or average used memory in MB
+  * @param maxHeap
+  *   \- current Max Heap size in MB
+  * @param usedOrAvgCpu
+  *   \- current or average cpu used in %
+  * @param maxStats
+  *   \- Option with tuple having max CPU and Memory consumption.
+  */
+case class CpuMemoryStats(
+  usedOrAvgMemory: Double,
+  maxHeap: Double,
+  usedOrAvgCpu: Double,
+  maxStats: Option[(Double, Double)]
+)
+
 object TimeMetric {
+
   val cal = Calendar.getInstance()
-  // Stage wise performance will be only recorded if parallel recording is started through initialize()
+
+  /** Stage wise performance will be only recorded if parallel recording is started through initialize(). This is
+    * accessible outside of this class, in case this information is required to be consumed outside of this class.
+    */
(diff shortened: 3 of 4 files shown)


codepropertygraph/src/main/scala/io/shiftleft/passes/CpgPass.scala
```diff
@@ -66,6 +67,7 @@ abstract class ForkJoinParallelCpgPass[T <: AnyRef](
     prefix: String = ""
   ): Unit = {
     baseLogger.info(s"Start of pass: $name")
+    TimeMetric.initiateNewStage(getClass.getSimpleName, Some(name), "ForkJoinParallelCpgPass")
```
DavidBakerEffendi, line 70: Would something like the following achieve the same?
```suggestion
    TimeMetric.initiateNewStage(getClass.getSimpleName, Some(name), getClass.getName)
```

codepropertygraph/src/main/scala/io/shiftleft/utils/TimeMetric.scala
```diff
@@ -0,0 +1,349 @@
...
+
+import java.io.PrintWriter
+import java.lang.management.ManagementFactory
+import java.util
+import java.util.concurrent.ConcurrentLinkedQueue
+import java.util.{Calendar, Date}
+import scala.collection.mutable
+case class TimeMetricRecordConfig(resultFile: String = "cpuandmemoryusage.csv", recordFreq: Int = 60000)
+case class MaxPrintBuffers(MAX_FULLNAME_BUFFER: Int = 90, MAX_NAME_BUFFER: Int = 55, MAX_DONE_TIME_BUFFER: Int = 40)
+object TimeMetric {
```
DavidBakerEffendi, line 14: Could this be a `class` instead? It would help to containerize all these variables into a single object instance which can be garbage collected much more easily.

codepropertygraph/src/main/scala/io/shiftleft/utils/TimeMetric.scala
```diff
@@ -0,0 +1,349 @@
...
+import com.sun.management.OperatingSystemMXBean
+import org.slf4j.{Logger, LoggerFactory}
+
+import java.io.PrintWriter
+import java.lang.management.ManagementFactory
+import java.util
+import java.util.concurrent.ConcurrentLinkedQueue
+import java.util.{Calendar, Date}
+import scala.collection.mutable
+case class TimeMetricRecordConfig(resultFile: String = "cpuandmemoryusage.csv", recordFreq: Int = 60000)
```
DavidBakerEffendi, line 12: Just a nit on spacing between imports and classes

codepropertygraph/src/main/scala/io/shiftleft/utils/TimeMetric.scala
```diff
@@ -0,0 +1,349 @@
...
+  private lazy val maxEndingBuffer = {
+    if (fullName) {
+      maxBuffers.MAX_FULLNAME_BUFFER
+    } else {
+      maxBuffers.MAX_NAME_BUFFER
+    }
+  }
+  private val osBean = ManagementFactory.getOperatingSystemMXBean().asInstanceOf[OperatingSystemMXBean]
+
+  def initialize(
```
DavidBakerEffendi, line 56: If we make `TimeMetric` a class, this code could be in the constructor (which I suppose is simply in the class body)

codepropertygraph/src/main/scala/io/shiftleft/utils/TimeMetric.scala
```diff
@@ -0,0 +1,349 @@
...
+  val stagePerformance: mutable.Map[String, NumberProcessor] = mutable.Map()
+  private val start                                          = cal.getTime
+  private var subStageStartTime                              = start
+  private var newTime                                        = start
+  private var stageStartTime                                 = start
+  private val ACROSS_PROCESS_STAGE                           = "<ACROSS_ALL>"
+  private val STAGE_NOT_SET                                  = "<not set>"
+  private val STARTED                                        = "Started"
+  private val DONE                                           = "Done"
+  private var lastStageName                                  = STAGE_NOT_SET
```
DavidBakerEffendi, line 26: Could these not be a stack instead? Then one could instead check if the stack is not empty and have fewer mutable variables. Also, I feel these could be combined into something like
```scala
final case class Stage(name: String, substage: mutable.Stack[String])

class TimeMetric {
// ...
  private val stages = mutable.Stack.empty[Stage]
// ...
}
```

codepropertygraph/src/main/scala/io/shiftleft/utils/TimeMetric.scala
```diff
@@ -0,0 +1,349 @@
...
+  private var lastSubStageName                               = STAGE_NOT_SET
+  private var additionalMetaData                             = STAGE_NOT_SET
+  private var alsoLogConsole                                 = false
+  private val logger: (String, Boolean) => Unit = (msg, subStage) => {
+    if (alsoLogConsole && (!subStage || !supressSubstages))
```
DavidBakerEffendi, line 31: Based on how this is used, should this boolean not be called `printToConsole`? Or something

codepropertygraph/src/main/scala/io/shiftleft/utils/TimeMetric.scala
```diff
@@ -0,0 +1,349 @@
...
+        while (true) {
+          // This while loop will take care of writing all the accumulated events in the queue first.
+          while (!queue.isEmpty) writer.println(queue.poll)
+          val (usedMemory, maxHeap, cpuUsed, _) = getCurrentCpuMemoryStats()
+          stagePerformance(ACROSS_PROCESS_STAGE).process(cpuUsed, usedMemory)
+          if (lastStageName != STAGE_NOT_SET) then stagePerformance (s"+ $lastStageName").process(cpuUsed, usedMemory)
+          if (lastSubStageName != STAGE_NOT_SET)
+            then stagePerformance (s"-- $lastSubStageName").process(cpuUsed, usedMemory)
+          writer.println(s"${getNewTime()}, $lastStageName, $lastSubStageName, $additionalMetaData, ${String
+              .format("%.2f", cpuUsed)}, ${String.format("%.2f", usedMemory)}, ${String.format("%.2f", maxHeap)}")
```
DavidBakerEffendi, line 97: Just wondering about locales, this may make a decimal `0,2` instead of `0.2` for example, which may break the CSV. I've encountered this often with my system local setting. This can be worked around by doing `"%.2f".formatLocal(java.util.Locale.US, usedMemory) `, which forces using a period instead of a comma.

Alternatively, you could wrap every column value in quotes

codepropertygraph/src/main/scala/io/shiftleft/utils/TimeMetric.scala
```diff
@@ -0,0 +1,349 @@
...
+    stage: Option[String] = None
+  ): (Double, Double, Double, Option[(Double, Double)]) = {
+    stage match {
+      case Some(st) if stagePerformance.contains(st) =>
+        val numberProcessor     = stagePerformance(st)
+        val (avgCpu, avgMemory) = numberProcessor.getAverage
+        val (maxCpu, maxMemory) = numberProcessor.getMax
+        (avgMemory, Runtime.getRuntime().maxMemory() * 0.000001, avgCpu, Some(maxCpu, maxMemory))
+      case _ =>
+        (
```
DavidBakerEffendi, line 263: Could you create a case class instead of return all of this in a tuple, please? It would help make it more descriptive as to what each variable is

codepropertygraph/src/main/scala/io/shiftleft/utils/TimeMetric.scala
DavidBakerEffendi: May also help enforce provenance if we need

codepropertygraph/src/main/scala/io/shiftleft/utils/TimeMetric.scala
```diff
@@ -0,0 +1,349 @@
...
+        recorder.get.start()
+      case _ =>
+    }
+  }
+
+  def setSupressSubstagesFlag(flag: Boolean): Unit = supressSubstages = flag
+
+  private class TimeRecorder(timeMetricRecordConfig: TimeMetricRecordConfig) extends Thread {
+
+    val queue: util.Queue[String] = new ConcurrentLinkedQueue[String]()
```
DavidBakerEffendi, line 80: If `TimeMetric` becomes a class, would a recorder have multiple threads accessing the queue?

codepropertygraph/src/main/scala/io/shiftleft/utils/TimeMetric.scala
DavidBakerEffendi: Is it the case that multiple threads/passes are writing to this at the same time?

codepropertygraph/src/main/scala/io/shiftleft/utils/TimeMetric.scala
DavidBakerEffendi: It may make sense to set up a helper method for formatting the double strings, since they are done all over the code

DavidBakerEffendi (commented):
Good idea, @mpollmeier could you give this a look? I recall some basic metrics captured in ODB but not sure if it's done at this level to include memory usage etc

codepropertygraph/src/main/scala/io/shiftleft/passes/CpgPass.scala
mpollmeier: +1 - a constant `ForkJoinParallelCpgPass` is wrong and misleading here, no?

mpollmeier:
> Good idea, @mpollmeier could you give this a look? I recall some basic metrics captured in ODB but not sure if it's done at this level to include memory usage etc

From what I can see and think of we only collect very basic metrics like number of changes in overflowdb's (and flatgraph's) DiffGraphApplier. The problem with time recorders is that it's not comparable between machines - I still think it's useful to have though! I will have a read through here now.

codepropertygraph/src/main/scala/io/shiftleft/passes/ParallelCpgPass.scala
```diff
@@ -58,6 +58,7 @@ abstract class ConcurrentWriterCpgPass[T <: AnyRef](
   ): Unit = {
     import ConcurrentWriterCpgPass.producerQueueCapacity
     baseLogger.info(s"Start of enhancement: $name")
+    TimeMetric.initiateNewStage(getClass.getSimpleName, Some(name), "ConcurrentWriterCpgPass")
```
mpollmeier, line 61:
```suggestion
    TimeMetric.initiateNewStage(getClass.getSimpleName, Some(name), getClass.getName)
```
same as above

codepropertygraph/src/main/scala/io/shiftleft/utils/TimeMetric.scala
mpollmeier: +1 for what David said - currently this is global mutable state, which may be accessed and modified by concurrently running CpgPasses.
(yes, currently they run in sequence, but the idea is that we can parallelise independent ones)

mpollmeier (commented):
mostly +1 to what David said

It'd be useful to have a minimised integration test, e.g. run the `io.shiftleft.mypass.SamplePass` (or similar) and verify that the measurements work in principle.

codepropertygraph/src/main/scala/io/shiftleft/passes/CpgPass.scala
pandurangpatil: `getClass.getName` will return the implementation class name. In order to understand if the given pass is extending "ForkJoinParallelCpgPass" or  "ConcurrentWriterCpgPass" . But we can read the parent class names and incorporate that as well. I will make that change.

codepropertygraph/src/main/scala/io/shiftleft/utils/TimeMetric.scala
pandurangpatil: It is not expected to call this TimeMetric from multiple places. It will be called from two places.
1. Parent stages like AST Creation, Overlay or OSS Dataflows. Which I assume will be mostly called from the main thread. These parent stages are like groups of passes. 
2. Second place it will be called from the passes, before all the tasks are spawned. In principle from their base class either from `ForkJoinParallelCpgPass` or `ConcurrentWriterCpgPass`. It gets before all the tasks are spawned so it will get called again from the main thread.

I will add few more details in the description which will give more clarity on it.

codepropertygraph/src/main/scala/io/shiftleft/utils/TimeMetric.scala
pandurangpatil: The reason I made it SingleTon as this might needs to be accessed from Joern as well as from the projects extending the Joern. Creating a class will make it difficult to pass on that instance, as that will call for interface change in many places.  
That's what I feel. Singleton Object would be the best option in this scenario.

codepropertygraph/src/main/scala/io/shiftleft/utils/TimeMetric.scala
pandurangpatil: I guess a stack can be created. However, the use case that it caters to will be limited to one parent stage (group of passes), and individual passes. Introducing a stack to make it dynamic will make some logic more complicated. Let me give it a try and let you know if I face any challenges.

codepropertygraph/src/main/scala/io/shiftleft/utils/TimeMetric.scala
pandurangpatil: beyond 2 decimals values just become noise in the report. In fact, we can even exclude the decimals and only show the numbers as that makes more sense.

codepropertygraph/src/main/scala/io/shiftleft/utils/TimeMetric.scala
pandurangpatil: As per my previous comment, Object will be the more suitable option for `TimeMetric`

pandurangpatil:
> > Good idea, @mpollmeier could you give this a look? I recall some basic metrics captured in ODB but not sure if it's done at this level to include memory usage etc
> 
> From what I can see and think of we only collect very basic metrics like number of changes in overflowdb's (and flatgraph's) DiffGraphApplier. The problem with time recorders is that it's not comparable between machines - I still think it's useful to have though! I will have a read through here now.

@mpollmeier even though it's not comparable between machines. It will be really helpful to understand which pass requires optimisation based on this data from a given machine. Otherwise, we are clueless and look through code, make guesses or use the profiler.

codepropertygraph/src/main/scala/io/shiftleft/passes/CpgPass.scala
pandurangpatil: BTW I am already passing  `Some(name)` => `Some(getClass.getName)`

codepropertygraph/src/main/scala/io/shiftleft/utils/TimeMetric.scala
pandurangpatil: I need to change the entire logic to use the stack. It will take more time to test the flow while making the changes to ensure things are working. In that case, I would probably not use the stack, as the changes are well-tested with all the combinations.

codepropertygraph/src/main/scala/io/shiftleft/passes/CpgPass.scala
pandurangpatil: Done

codepropertygraph/src/main/scala/io/shiftleft/utils/TimeMetric.scala
pandurangpatil: done

codepropertygraph/src/main/scala/io/shiftleft/utils/TimeMetric.scala
pandurangpatil: done

codepropertygraph/src/main/scala/io/shiftleft/utils/TimeMetric.scala
pandurangpatil: done

codepropertygraph/src/main/scala/io/shiftleft/utils/TimeMetric.scala
pandurangpatil: done

codepropertygraph/src/main/scala/io/shiftleft/passes/ParallelCpgPass.scala
pandurangpatil: done

pandurangpatil:
> mostly +1 to what David said
> 
> It'd be useful to have a minimised integration test, e.g. run the `io.shiftleft.mypass.SamplePass` (or similar) and verify that the measurements work in principle.

Unit test has been added.

pandurangpatil:
@DavidBakerEffendi @mpollmeier I have added comprehensive documentation to the file which will try to explain the implementation. Let me know if you have further questions. 

I have also addressed the majority of review comments in the code. Where I think it's difficult to make the changes, I have replied to the comment.

pandurangpatil:
>+1 for what David said - currently this is global mutable state, which may be accessed and modified by concurrently running CpgPasses.
(yes, currently they run in sequence, but the idea is that we can parallelise independent ones)

Probably I am not able to connect how making Class we can achieve object not being accessed concurrently. If we have to write the status to the common file, we will have to access the file concurrently.

codepropertygraph/src/main/scala/io/shiftleft/utils/TimeMetric.scala
DavidBakerEffendi: @mpollmeier What do you think here?

pandurangpatil:
@DavidBakerEffendi @mpollmeier 

We have to record the data from start to end.

we need to make sure we initialise it outside any of the passes and end it towards the end when everything is done. 

As well as we should be able to start a stage from passes and end the respective stage there. We need to have something like Singleton solution. Otherwise, we have to pass on the object through every pass created in Joern.

mpollmeier:
>  We need to have something like Singleton solution. Otherwise, we have to pass on the object through every pass created in Joern.

It may well be very complicated, and I'm open for a singleton, but I'd like to at least give it a try via argument passing. Most passes are invoked from a few places, so passing the TimeMetric instance through (via regular or implicit argument) might not be as bad as it sounds. But that's only my gut feeling, did you give it a try yet?

codepropertygraph/src/main/scala/io/shiftleft/utils/TimeMetric.scala
mpollmeier: responded in a comment below: https://github.com/ShiftLeftSecurity/codepropertygraph/pull/1759#issuecomment-1991185785

DavidBakerEffendi:
@pandurangpatil @mpollmeier Copying from https://discord.com/channels/832209896089976854/1215216562198290483/1216974330630639666

In finding a balance between static object and classes, let me propose that `TimeMetric` has both an object and class component, but that object `TimeMetric` keeps track of all the recorders in flight. 

When a pass starts, all passes are given a `Cpg`, where the graph can be used as uniquely identifying a pass to a recorder. e.g. when a pass starts, it can check
Am I recording?
If yes, `TimeMetric.pushStage(..., cpg.graph)`

Then object `TimeMetric` could have some map of `mutable.TrieMap[Graph, TimeMetric]` so if a user is recording multiple CPG generations at the same time we won't modify the same global recorder

pandurangpatil:
@mpollmeier @DavidBakerEffendi I acknowledge both the suggestion. Let me give it a try and update here.

pandurangpatil:
@DavidBakerEffendi as discussed with one urgent requirement, I need to get this matrix logger deployed to the customer end to collect the precise data points in the absence of code access. So created this temporary arrangement with this PR - https://github.com/ShiftLeftSecurity/codepropertygraph/pull/1765 to intercept the starting and ending events. 

I will work on finalising this PR as soon as I get free from the priority items.

cc: @mpollmeier

PR #1762, opened 2024-03-21
Added catch and finally to ControlStructureTypes

max-leuthaeuser:


max-leuthaeuser:
Hm fails even though I ran generateDomainClasses and committed everything. Can somebody fix this PR?

PR #1765, opened 2024-04-09
Stats logger interceptor

pandurangpatil:
Temporary mechanism to intercept the Passes and capture starting and ending events.

### DIFF
--- a/codepropertygraph/src/main/scala/io/shiftleft/utils/StatsCollector.scala
+++ b/codepropertygraph/src/main/scala/io/shiftleft/utils/StatsCollector.scala
@@ -1,0 +1,28 @@
+package io.shiftleft.utils
+
+object StatsLogger extends DataLogger {
+  private var logger: Option[DataLogger] = None
+  def initialise(logger: Option[DataLogger] = None): Unit = {
+    this.logger = logger
+  }
+
+  def initiateNewStage(
+    stageName: String,
+    stageFullName: Option[String] = None,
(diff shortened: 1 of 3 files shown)


DavidBakerEffendi (approved):

mpollmeier (changes requested):
please don't merge yet, I'll add some comments in a bit

mpollmeier:
Having the logger as a global object with mutable state seems like a bad idea, e.g. because we want to be able to run passes in parallel. 
Also, the `initialise` method isn't ever called, not sure how and if this even works...?

codepropertygraph/src/main/scala/io/shiftleft/utils/StatsCollector.scala
```diff
@@ -0,0 +1,28 @@
+package io.shiftleft.utils
+
+object StatsLogger extends DataLogger {
```
DavidBakerEffendi, line 3: Why are we making this an object instead of a class that we can keep all fields immutable?

pandurangpatil:
@mpollmeier @DavidBakerEffendi this is an temporary arrangement I am making to get some urgent troubleshooting done on customer on premise setup. As I don't have access to customer code, I need to get this one pushed for time being quickly. I will make sure in few days to update the original PR which will address the previous PR comments.

DavidBakerEffendi:
@pandurangpatil Would it not make sense to then set up a local release on the customer's side? Or perhaps have a release for a Privado fork on this to benefit you guys in these situations long term?

PR #1771, opened 2024-06-25
use scala primitives (e.g. Integer -> Int)

mpollmeier:
motivation: minify diff for flatgraph migration

changes re Properties.scala in generated code: https://github.com/ShiftLeftSecurity/codepropertygraph/pull/1771/files#diff-9e678c672157f7a6104f7cf5995958f319155b18d435f82fe7b053e2433f0bd1


bbrehm (approved):

mpollmeier:
argh. GH showed green lights here, even though the pr run was red. 
might just work though, since now the codegen should be published

PR #1782, opened 2024-08-02
markus/removeTestcoded

ml86:
Remove unused test code files.


codepropertygraph/src/main/scala/io/shiftleft/SerializedCpg.scala
```diff
@@ -7,6 +7,7 @@ import java.net.{URI, URISyntaxException}
 import java.nio.file.{FileSystem, FileSystems, Files}
 import java.util
 
+@deprecated
```
mpollmeier, line 10: a comment re what to use instead is always welcome on deprecated messages.

mpollmeier (approved):

codepropertygraph/src/main/scala/io/shiftleft/SerializedCpg.scala
ml86: I accidentally added this change to the PR. Will address this in another one.

PR #1785, opened 2024-09-12
update flatgraph

maltek:
also make updateDependencies.sh compatible with non-GNU sed variants


mpollmeier (approved):
🤦🏻 the sed gala continues, plot worsening

PR #1789, opened 2024-11-05
sonatype central

mpollmeier:
We migrated to sonatype central due to publishing problems, as
suggested by their support. Let's see if this works.

https://central.sonatype.com/


mpollmeier:
actually no, this is not published under joernio...

PR #1809, opened 2025-02-05
Centralized all protoIds into `ProtoIds`

DavidBakerEffendi:
As a means of record keeping, aggregated all ProtoIds entries into a single object.

Many items seem to collide, however they are often part of different schema objects.

To reviewers: Seems that uniqueness is a pair consisting of the ProtoId + Schema Object Type. E.g., ID + ConstantCategory or ID + Node, etc. One may choose to instead group the ProtoIds following this composite-style uniqueness instead of in a single object.


ml86 (approved):
@DavidBakerEffendi I think an unstructured listing like it is now is fine. Otherwise you would now have to figure out the logical groups in which an ID needs to be unique. We also would need to maintain this grouping if we add new elements.

PR #1815, opened 2025-05-07
log node/edge/property count when loading proto CPGs

maltek:


mpollmeier (changes requested):
https://github.com/joernio/flatgraph/pull/320#pullrequestreview-2821843770 applies here as well

maltek:
I've added logging of counts during deserialization over in the other PR, so this PR now only logs counts for proto CPGs (the only CPGs not going through that code path)

mpollmeier (approved):

PR #1819, opened 2025-05-21
Delete Location and NewLocation

Flandini:
In https://github.com/joernio/joern/pull/5482, we are switching to lazily computed location info. We are replacing the `Location` and `NewLocation` types with types defined in the `joern` repo, and the location info does not need to be in the cpg.


Flandini:
These should also be deleted from the cpg spec website. It looks like this is an upload of a `schema.json` file generated from this repo?

ml86 (approved):

PR #1827, opened 2025-07-10
loneElement and loneElementOption: extension methods for Iterable

mpollmeier:


codepropertygraph/src/main/scala/io/shiftleft/Implicits.scala
```diff
@@ -3,45 +3,85 @@ package io.shiftleft
...
+    /** @see {{{loneElementOption(hint)}}} */
+    def loneElementOption: Option[A] =
+      loneElementOption(hint = None)
+
+    /** @return
+      *   {{{Some(element)}}} if the Iterable has exactly one element, or {{{None}}} if the Iterable has zero or more
+      *   than 1 element. Note: if the lone element is {{{null}}}, this will return {{{Some(null)}}}, which is in
+      *   accordance with how {{{headOption}}} works.
+      */
+    def loneElementOption(hint: String | None.type = None): Option[A] = {
```
maltek, line 55: the hint doesn't seem to be used anywhere? Not sure what we should do with it either - log something?

mpollmeier: oh wait, forgot to refactor that one to the end

maltek (approved):

mpollmeier:
as discussed via slack, will move this downstream into flatgraph, since it's used there most often

maltek (approved):

PR #1856, opened 2026-06-02
Change INDEX property default value to -2.

ml86:
Previously INDEX property default value was -1 now it is -2.
We do this to avoid representation collisions if frontends happen to not
set this property. See code comment for further details.

### DIFF
--- a/schema/src/main/scala/io/shiftleft/codepropertygraph/schema/Base.scala
+++ b/schema/src/main/scala/io/shiftleft/codepropertygraph/schema/Base.scala
@@ -74,7 +74,7 @@ object Base extends SchemaBase {
                     |""".stripMargin
       )
       // We use a default value that is different from -1 because some code working with parameters and return
-      // values, represent then in a single number plane where -1 encodes the return value and we do not want
+      // values represent them in a single number plane where -1 encodes the return value and we do not want
       // the "value unset" default value to collide with that.
       .mandatory(-2)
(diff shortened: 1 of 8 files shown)


schema/src/main/scala/io/shiftleft/codepropertygraph/schema/Base.scala
```diff
@@ -73,7 +73,10 @@ object Base extends SchemaBase {
                     | self / this parameter.
                     |""".stripMargin
       )
-      .mandatory(PropertyDefaults.Int)
+      // We use a default value that is different from -1 because some code working with parameters and return
+      // values, represent then in a single number plane where -1 encodes the return value and we do not want
```
maltek:
```suggestion
      // values represent them in a single number plane where -1 encodes the return value and we do not want
```

maltek (approved):

SuperUserDone (approved):

SuperUserDone:
Might need to have a look here: https://github.com/ShiftLeftSecurity/codescience/blob/master/semanticcpg-ext/src/main/scala/io/shiftleft/semanticcpgext/passes/varargparamindex/VarArgParameterPass.scala

ml86:
@SuperUserDone The `VarArgParameterPass` is fine. The `-1` encoding of "method has no vararg parameter is independent of the INDEX default value.

PR #1859, opened 2026-07-06
Add bazel build.

ml86:


MODULE.bazel
```diff
@@ -0,0 +1,77 @@
...
+bazel_dep(name = "buildifier_prebuilt", version = "8.5.1.2")
+
+JOERN_VERSION = "db325972fdf72e53f76beba77b72e84b2c469905"
+
+git_override(
+    module_name = "bazel_tooling",
+    commit = "{}".format(JOERN_VERSION),
+    remote = "git@github.com:joernio/joern.git",
+    strip_prefix = "bazel/tooling",
+)
```
maltek, line 16: It would be nice to have this in a separate repo, for less circular importing. But not a must for me.

maltek (approved):

MODULE.bazel
ml86: The plan is to eventually bring codepropertygraph into the Joern repo which would resolve this. In the mean time i try to avoid even more repositories.

